Issue #219, opened 2024-07-01
`jtot_hires` is not always needed


theo-brown:
`jtot_hires` is meant to be used when computing the psi field from prescribed currents.

As this is only done in some cases, `jtot_hires` should be an optional parameter of the `Currents` dataclass.

For example, the function [`_calculate_currents_from_psi`](https://github.com/google-deepmind/torax/blob/c1d674f55eb2c812c079b579116e1b945254dc7d/torax/core_profile_setters.py#L384) is meant to compute self-consistent currents from the psi field, and therefore does not require the `jtot_hires` to be set.

PR #220, opened 2024-07-01
Make jtot_hires optional

theo-brown:
Fixes issue #219.

Note that although the existing tests pass, I have not added any new tests to check this functionality. However, this should not be a problem, as the bit I removed was permanently unused anyway.

### DIFF
--- a/torax/core_profile_setters.py
+++ b/torax/core_profile_setters.py
@@ -391,7 +391,7 @@ def _calculate_currents_from_psi(
     psi: cell_variable.CellVariable,
     source_models: source_models_lib.SourceModels,
 ) -> state.Currents:
-  """Creates the initial Currents using psi to calculate jtot.
+  """Creates the initial currents using psi to calculate jtot.
 
   Args:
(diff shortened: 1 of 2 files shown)


theo-brown:
One thing I wasn't certain of was whether setting a frozen Chex dataclass field to Optional was permitted.

torax/core_profile_setters.py
```diff
@@ -391,7 +391,7 @@ def _calculate_currents_from_psi(
     psi: cell_variable.CellVariable,
     source_models: source_models_lib.SourceModels,
 ) -> state.Currents:
-  """Creates the initial Currents using psi to calculate jtot.
```
jcitrin, line 394: currents can remain capitalized since it refers to the class

jcitrin:
> One thing I wasn't certain of was whether setting a frozen Chex dataclass field to Optional was permitted.

Shouldn't be an issue. See other parts of TORAX where we do this, e.g. https://github.com/google-deepmind/torax/blob/b13133f9a3bd4a4a8fd3ab4ba71949244453a67c/torax/fvm/cell_variable.py#L67

Issue #212, opened 2024-06-28
[docs] `PRESCRIBED` heating and current source profiles


theo-brown:
While future work may focus on developing specific physically realistic profiles for current and heating actuators, in the meantime it would be useful to be able to **represent arbitrary source profiles** so that scenarios generated in other codes can be run in TORAX.

The most basic of these would be a piecewise linear profile.

theo-brown:
I would be happy to extend `sources/formulas.py` and `sources/formula_config.py` to support this, if a core contributor would give me the go-ahead! Not sure if there is a better option that I'm missing.

jcitrin:
Fully agree that supporting arbitrary source profiles should be done in the near-term. 

At the moment we are working on supporting arbitrary initial and prescribed core profiles, and the infrastructure is already there to a large extent. `Te`, `Ti`, `ne` are now of type `TimeInterpolatedArray` (see `interpolated_param.py`) and can be defined in config as nested dicts in a time series `{time: value}` like `{0: {0.0: 20, 0.95: 5, 1.0: 0.2}, 1: {0.0: 20, 0.95: 5, 1.0: 0.2}}`, where value is itself a dict representing `{rhonorm: value}`. Down the road, we plan to generalize `TimeInterpolatedArray` to accept more general datastructures such as xarray datasets or numpy arrays, as long as they satisfy specific formats. Work has started in this direction, see #197 . This will enable loading these values from output files of previous runs, from IMAS (there are IMAS IDS to xarray converters out there), or more generally from any userdata as long as it's converted to the right format.

After this long preamble :) , I think we should reuse all this infrastructure to enable arbitrary source profiles as well. Can envisage adding a new supported mode (see `Mode` class in `sources/runtime_params.py`) like `PRESCRIBED`, and the source then gets as input the nested dict, or numpy arrays, or an xarray dataset, or a filepath to an .nc file or IMAS h5 file, etc (note that currently only nested dict input is supported). The `PRESCRIBED` mode could include options like a multiplication factor to the input, or normalizing the input to a desired total integrated source amplitude.

You could indeed also extend `sources/formulas.py` and `sources/formula_config.py` to do the same. This is easier since there's less new development to do, but likely down the road we'll move this into a new `PRESCRIBED` mode. I'd recommend a new "formula" class something like (just a suggestion)

```
@dataclasses.dataclass
class PiecewiseLinear:
  """Configures a piecewise linear profile."""

  use_total: bool = False
[...]

PR #221, opened 2024-07-01
Add support for prescribed source profiles

theo-brown:
Pull request for #212. 

Plan is to use `TimeInterpolatedArray` for prescribing potentially time-evolving source profiles.

Key files are `torax/sources/source.py` and `torax/sources/runtime_params.py`.


theo-brown:
Progress stalled pending #222 .

torax/sources/runtime_params.py
```diff
@@ -104,3 +120,5 @@ class DynamicRuntimeParams:
   mode: int
   is_explicit: bool
   formula: formula_config.DynamicFormula
+  # Prescribed values on the face grid.
+  prescribed_values: chex.Array
```
araju, line 124: For most places in TORAX, if they don't have "face" in the var name, it refers to the cell grid. For this variable, can you rename it to `prescribed_face_values` or `prescribed_values_face_grid` or something along those lines?

torax/sources/source.py
```diff
@@ -441,12 +446,15 @@ def get_source_profiles(
     core_profiles: state.CoreProfiles | None,
     model_func: SourceProfileFunction,
     formula: SourceProfileFunction,
+    prescribed_values: chex.Array | None,
```
araju, line 449: I think the function below would fail if `prescribed_values` was `None`. And I dont think you ever pass in None, anyways. So go ahead and just have this be a `chex.Array`. Also, same point on naming from the other comment - add "face" to the name.

torax/sources/tests/source.py
```diff
@@ -252,10 +252,38 @@ def test_defaults_output_zeros(self):
           profile,
           source_lib.ProfileType.CELL.get_zero_profile(geo),
       )
+    with self.subTest("prescribed"):
```
araju, line 255: nit - use single-quotes because that's whats in the rest of the file, for consistency.

torax/sources/external_current_source.py
```diff
@@ -227,16 +231,28 @@ def jext_hires(
     """Return the external current density profile along the hires cell grid."""
     assert isinstance(dynamic_source_runtime_params, DynamicRuntimeParams)
     self.check_mode(dynamic_source_runtime_params.mode)
+
+    # Interpolate prescribed values onto the hires grid
+    if dynamic_source_runtime_params.mode == runtime_params_lib.Mode.PRESCRIBED:
+      hires_prescribed_values = jnp.interp(
+        geo.r_hires_norm,
+        geo.r_norm,
+        dynamic_source_runtime_params.prescribed_values
```
araju, line 240: this code block is actually not jax-compatible, meaning if you were to change the `Mode` to/from prescribed, the original code path would still be followed.

use `jnp.where` instead of an if-else

araju (changes requested):
This looks great! Sorry for the delay on reviews. Main thing you want to do is rebase to head because some of the sources API changed recently I think.

Once you update these bits, I'll take another quick look before approving.

Thanks for working on this. This is a valuable feature in TORAX.

torax/sources/runtime_params.py
theo-brown: Cool, that's good to know generally!

theo-brown:
Rebased onto main ef5a49d in fbb0837

torax/sources/source.py
theo-brown: Good point about the array.

theo-brown: I think this one actually should be cell rather than face - the `formulas` and `models` both compute source terms on the cell grid, and before `get_source_profiles` is called the `prescribed_values_face` (on the right hand side) are interpolated onto the cell grid (left hand side) [[link](https://github.com/theo-brown/torax/blob/fbb0837ba2693bcd645e39a406efc95ea3f3cdfe/torax/sources/source.py#L221)].

theo-brown: Actually, after doing a bit more digging, it might not be that simple. `get_source_profiles` is used to compute profiles on both the cell and face grids:

[Cell example, `fusion_heat_source.py`](https://github.com/google-deepmind/torax/blob/ef5a49ddf87de18dabc172ae57fd8fec4f6f2e89/torax/sources/fusion_heat_source.py#L118)
[Face example, `external_current_source.py`](https://github.com/google-deepmind/torax/blob/ef5a49ddf87de18dabc172ae57fd8fec4f6f2e89/torax/sources/external_current_source.py#L208C22-L208C41)

theo-brown: Not sure whether this is intentional/something that should be continued, or whether it's a quirk of the `jext` stuff. There are a few things that are only used in `jext` that are not used elsewhere. Is `jext` also due a refactor, to make it more flexible?

torax/sources/runtime_params.py
jcitrin: Since prescribed_values uses the TimeInterpolated patterns, then on the config level it's not necessarily on either the face or cell grid, it's on a user-provided grid that then gets interpolated onto the TORAX grids. So the issue I think is actually with the comment rather than the variable name

torax/sources/source.py
jcitrin: See comment above. Unless I'm confused, the input into `prescribed_values` on the config level can be on any user-provided grid (as long as it contains rhon=0 and rhon=1), and the `get_value` eventually provides the face grid interpolation that can then easily also be converted to the cell grid

torax/tests/test_data/test_prescribed_particle_source.py
```diff
@@ -0,0 +1,51 @@
...
+        'geometry_type': 'circular',
+    },
+    'sources': {
+        'generic_ion_el_heat_source': {},
+        'qei_source': {},
+        'gas_puff_source': {
+            'S_puff_tot': 1.0e22,
+        },
+        'pellet_source': {
+            'S_pellet_tot': {0: 5.0e22, 2.5: 5.0e22, 2.6: 0.5e22},
```
jcitrin, line 32: I'm not sure that this is testing the new PRESCRIBED mode, since this uses preexisting features

theo-brown: You're absolutely right, I'll change it.

theo-brown: Changed to prescribing jext in 0c8d5e693db052bc3c475fa0dab00613c90e4800

torax/sources/runtime_params.py
theo-brown: Comment removed in a50e64bded6af7913c40a13bb9179a2e58c2306b

torax/sources/source.py
theo-brown: Based on @jcitrin's comment, I'm leaving it as `prescribed_values` unless there are any further objections.

torax/sources/external_current_source.py
theo-brown: Resolved in 8e32310c1220e9316e38dfe54c380581aec388d8

torax/sources/tests/source.py
theo-brown: Resolved in 8cd2393379ebaed4ed94ac327d1bbffd65c3c4d1

torax/sources/external_current_source.py
```diff
@@ -228,17 +232,30 @@ def jext_hires(
...
-        geo=geo,
-        core_profiles=None,
-        # There is no model for this source.
-        model_func=(lambda _0, _1, _2, _3: jnp.zeros_like(geo.r_hires_norm)),
-        formula=self.hires_formula,
-        output_shape=geo.r_hires_norm.shape,
+
+    # Interpolate prescribed values onto the hires grid
+    hires_prescribed_values = jnp.where(
+      dynamic_source_runtime_params.mode == runtime_params_lib.Mode.PRESCRIBED,
```
araju, line 238: Check against the value of the enum, not the enum itself, like shown here: https://github.com/google-deepmind/torax/blob/bb08d5a432ad94235042d7b107c8e6314b00389d/torax/sources/source.py#L471

Python will end up comparing an enum value vs an int and always return false.

torax/sources/runtime_params.py
araju: Ahh, great point. Thanks for catching that Jonathan.

torax/sources/source.py
```diff
@@ -216,13 +217,20 @@ def get_value(
         if self.formula is None
         else self.formula
     )
+    # dynamic_source_runtime_params.prescribed_values will be an array
+    # interpolated on the face grid (see config.config_args.interpolate_var_2d)
+    # TODO: This is non-robust, as it assumes the output shape is the cell grid.
```
araju, line 222: minor nit - just for our own internal linting, this TODO will cause some issues. Can you open up a discussion or issue regarding this, and then we can file it away properly as a TODO?

Here, can you make this a NOTE that this assumes output on a cell grid.

araju (approved):

torax/sources/source.py
theo-brown: Ok! I'll add it as a discussion when the PR's been merged.

torax/sources/external_current_source.py
theo-brown: Resolved in 95ae88f699bac31839c0529f4529a3c2dfdd7801

araju:
I cleaned up a couple small things in this PR in our internal system, and need one more internal approval before I can merge this thing.

araju:
Just so you're aware - this PR is not passing all our internal tests yet when merged with HEAD. I'm not sure why and have been debugging, but this is what's blocking things now. Will try and keep you posted.

theo-brown:
Is there anything I can do to help? I had thought it had passed the relevant tests (using `pytest`) in the repo, is it failing on those or is it some internal secret (non-open) test?

araju:
> Is there anything I can do to help? I had thought it had passed the relevant tests (using `pytest`) in the repo, is it failing on those or is it some internal secret (non-open) test?

I was confused here as well. I had made very minor cosmetic changes on top, so I thought it was something I did.

But actually, I just checked out your fork and your branch and tested, and I'm getting the same errors. My guess is the rebase broke something in your code.

I'm looking into it as well, but if you could check out your branch again and run `pytest torax/tests/sim.py` you'll see the errors I'm getting as well.

theo-brown:
Following this trail of `jext` problems, I've just checked the simulations that are being produced.
This is the simulation from `test_prescribed_jext.py`, where the `jext` is smoothly interpolated between two Gaussians.

**t=0s + a bit to avoid initial condition transients**
image

**t=5s**
image

I'm concerned that I've missed a `jext` update somewhere, as the `jtot` and `johm` profiles seem to keep the lump from the original `jext` profile rather than it moving around as `jext` is updated.

theo-brown:
Given that this particular chain of problems is to do with `jext`, I thought I should do a sense-check on the simulations produced and have a look at their `jext` profiles.

This is the simulation from `test_prescribed_jext.py`, where the `jext` is smoothly interpolated between two Gaussians. In the last commit I updated the data for this test so that the test passes.

**t=0s + a bit to avoid initial condition transients**
image

**t=5s**
image

I'm concerned that I've missed a `jext` update somewhere, as the `jtot` and `johm` profiles seem to keep the lump from the original `jext` profile rather than it moving around as `jext` is updated. I'll admit to not really having an understanding of what should be happening physically here. @jcitrin would appreciate your input as to whether this particular sim is producing reasonable results.

araju:
Awesome! Nice catch. That definitely fixed the test you added. You may have already seen this, but 3 other test cases from `tests/sim.py` are stil not passing:

 - test_torax_sim_test_iterhybrid_rampup
 - test_torax_sim_test_prescribed_timedependent_ne
 - test_torax_sim_test_timedependence

I didn't get a chance to find the underlying issue yesterday but am looking again today.

I'll let @jcitrin comment on the other questions you had.

theo-brown:
## Problem 2: code in this PR for `jext` calculation breaks examples where `jext` is a fixed fraction of the total power

The PR is failing existing integration tests.
The tests in question are:

- test_prescribed_timedependent_ne
- test_timedependence
- test_iterhybrid_rampup

Plotting the simulation result for main (1, solid) vs this PR (2, dashed) gives the following for the `test_iterhybrid_rampup` scenario:

image

As you can see, this PR's solution has the `jext` profile (purple, bottom left) varying over time, rather than being fixed. This is because the jext is being recalculated at every timestep, and the jext formula scales the jext profile to be a fixed fraction of the total current.

So: there needs to be a way of setting jext to be a fixed fraction of jtot at t=0, and then freezing that jext profile for all time.

fcasson:
> I'm concerned that I've missed a jext update somewhere, as the jtot and johm profiles seem to keep the lump from the original jext profile rather than it moving around as jext is updated. I'll admit to not really having an understanding of what should be happening physically here

If you add a external CD, instantaneously nothing changes in j_tot.  Because of faradays law, the J_in should exactly oppose what you added.  Then you wait for J_in to decay away for the current drive to be realised.  So it's not obviously wrong to me, changes in j_tot take time to show up.  j_in should change rapidly when you change j_ext, but not j_tot

theo-brown:
Also note that all of these difficulties are actually to do with jext, rather than with the prescribed source profiles. It could be worthwhile splitting this PR up so that the prescribed profiles bit goes in one PR and the jext bit goes in another.

jcitrin:
> As you can see, this PR's solution has the jext profile (purple, bottom left) varying over time, rather than being fixed. This is because the jext is being recalculated at every timestep, and the jext formula scales the jext profile to be a fixed fraction of the total current. So: there needs to be a way of setting jext to be a fixed fraction of jtot at t=0, and then freezing that jext profile for all time.

@araju and I just looked into this in detail.

What's happening is that you fixed a bug that we didn't realize we had, thanks! :)

The default runtime_params for `jext` is `use_absolute_jext==False`, meaning that the `jext` amplitude is set as a Greenwald fraction according to `fext`. In the "failing cases", the behaviour in your PR in that `jext` is time-dependent, is correct, since `Ip` is changing and thus the Greenwald factor is changing. These changes are picked up in your PR in `source_models.py` due to the `get_values()` addition

image

Previously, `jext` would come from `core_profiles.currents` (for legacy reasons that will be now removed), and it turns out that we neglected to update `jext` in `core_profiles.currents` for time-dependent cases, so it was frozen at the initial condition.

jcitrin:
So we will first fix this bug and do some cleanup, and then rebase your PR on top of that and fix any merge conflicts.

Regarding Problem #1 , yes as Francis says, it takes a long time for the current to diffuse for these parameters, so that `johm` bump will persist. For a sanity check, you can redo the case with a large `resistivity_mult` in the `numerics` dict, e.g. 100, which will decrease the current diffusion time and you should see faster `johm` and `jtot` dynamics

araju:
What we're going to do is fix this bug at head, and then we can merge you PR as well right after. You don't need to do anything on this anymore.

I know you might be on vacation for a bit soon. Just in case, do you mind making sure I have push access to this branch in your forked repo? Just so I can push any changes that may need to happen to resolve conflicts. I _should_ be able to handle conflicts and merge the PR without you, but that is just a precaution.

Thanks again for uncovering this!

theo-brown:
Amazing! Sounds like it's a good result all round. I've sent @araju an invite to be a collaborator on my fork.
Thanks for your time in handling this PR, everyone!

araju:
Hey Theo, sorry for the delay with all of this. Lost a week or so with a cold, so thanks for bearing with me!

Not sure why the commit we merged separately didnt sit on top of this PR. We're still getting the hand of our internal tools with dealing with the Github PRs haha. But we can close this now that [6aeb322](https://github.com/google-deepmind/torax/commit/6aeb3223223fecce9af1d0d8dcd90a4ac624851a) is submitted.

Issue #373, opened 2024-08-30
EQDSK support for geometry


fusionby2030:
TORAX has a lot of potential and I have been excited to use it. However,  I believe an outstanding issue is the lack of generalized geometry support. To this end, I would like to help add EQDSK support. Many EQDSK reading and writing tools exist already (most derived from Ben Dudson, c.f. [TSVV-15's](https://github.com/Fusion-Power-Plant-Framework/eqdsk) or [freegs](https://github.com/freegs-plasma/FreeQDSK)), which are able to handle the COCOS transformations given an g/f/-EQDSK file. The main trick is taking equilibrium from EQDSK and calculating the flux surface averaged quantities. To this end, I append the following code which does this and compare it to a CHEASE run. 

```
from eqdsk import EQDSKInterface # https://github.com/Fusion-Power-Plant-Framework/eqdsk
from scipy.integrate import cumulative_trapezoid
from scipy.interpolate import interp1d, RectBivariateSpline
from contourpy import contour_generator 
import numpy as np 

eq_fpath = './eqdsk_cocos02.eqdsk'
eq       = EQDSKInterface.from_file(eq_fpath, no_cocos=True)
eqfile   = eq.__dict__

vaccum_permativity = 1.25663706212e-6

# ---- Scalars 
RMAJ = eqfile['xmag']
RMIN = eqfile['xbdry'].max() - RMAJ
B0   = eqfile['bcentre']
# ---- Grids
# Psi Uniform Grid
psi_eqdsk_1dgrid = np.linspace(eqfile['psimag'], eqfile['psibdry'], eqfile['nx'])

# X and Z 1D grid, also Meshgrid
X_1D = np.linspace(eqfile['xgrid1'], eqfile['xgrid1'] + eqfile['xdim'], eqfile['nx'])
Z_1D = np.linspace(eqfile['zmid'] - eqfile['zdim']/2, eqfile['zmid'] + eqfile['zdim']/2, eqfile['nz'])
X, Z = np.meshgrid(X_1D, Z_1D, indexing='ij')

# Plasma Bndry (Last closed flux surface)
Xlcfs, Zlcfs = eqfile['xbdry'], eqfile['zbdry']

# Psi 2D grid defined on the Meshgrid
psi_eqdsk_2dgrid  = eqfile['psi']
# normalised Psi w.r.t boundary
[...]

PR #391, opened 2024-09-08
Adding EQDSK support 

fusionby2030:
A draft for EQDSK support by modifying the geometry, geometry_loader, and build_sim files. 

A corresponding geometry file is contributed to `data/third_party/geo` (credit to Antti Salmi for the help in geometry file generations). 

Additionally, an example configuration is made based off  (copied from)  the iter hybrid scenario and basic example already provided.  But the PR can likely do without the `eqdsk_example` config... the example takes a long time to run (at least it runs :D)...

The flux surface averaging is done using contourpy and RectBivariate spline interpolation of the PSI 2D grid. Note, there appear to be systematic differences in the shaping parameters (R inboard, R outboard, triangularity).

Somewhere the requirements have to be updated to include the [TSVV-15's eqdsk](https://github.com/Fusion-Power-Plant-Framework/eqdsk) and [contourpy](https://pypi.org/project/contourpy/)

See #373 for more details and discussion.


torax/geometry.py
```diff
@@ -28,6 +28,7 @@
 import jax.numpy as jnp
 import numpy as np
 import scipy
+from contourpy import contour_generator
```
tamaranorman, line 31: Can you add new dependencies to here

https://github.com/google-deepmind/torax/blob/main/pyproject.toml

Also can you make it such that the dependencies are ordered alphabetically - although keeping the groupings

torax/geometry.py
```diff
@@ -879,6 +880,213 @@ def from_fbt(
...
+  def from_eqdsk(
+        cls,
+        geometry_dir: str | None = None,
+        geometry_file: str = 'eqdsk_cocos02.eqdsk',
+        hires_fac: int = 4,
+        Ip_from_parameters: bool = True,
+        n_rho: int = 25,
+        # NOTE: The other geometry factors are coming from eqdsk...
+    ):
+        # helper function to calculate area of a polygon given curve along x, z
```
tamaranorman, line 893: Use """ method comments for method level comments

And also add types

torax/geometry.py
```diff
@@ -879,6 +880,213 @@ def from_fbt(
...
+        def calculate_area(x, z):
+            # Gauss-shoelace formula (https://en.wikipedia.org/wiki/Shoelace_formula)
+            # could likely use the determinant approach
+            #  0.5 * np.abs(np.dot(x, np.roll(z, 1)) - np.dot(z, np.roll(x, 1)))
```
tamaranorman, line 897: Is this an alternative method, if we aren't using this I am not sure we need to add this comment

torax/geometry_loader.py
```diff
@@ -58,6 +59,11 @@ def _load_fbt_data(file_path: str) -> dict[str, np.ndarray]:
   """Loads the data from a FBT-LY file into a dictionary."""
   return scipy.io.loadmat(file_path, squeeze_me=True)
 
+from eqdsk import EQDSKInterface
```
tamaranorman, line 62: Move imports to the top and also add to dependency file

torax/geometry_loader.py
```diff
@@ -58,6 +59,11 @@ def _load_fbt_data(file_path: str) -> dict[str, np.ndarray]:
...
+def _load_eqdsk_data(file_path: str) -> dict[str, np.ndarray]:
+    eq = EQDSKInterface.from_file(file_path, no_cocos=True) # should probably handle COCOS shenanigans...
```
tamaranorman, line 64: Is this something we need to handle here? Or maybe turn it into a formal todo

torax/geometry_loader.py
```diff
@@ -58,6 +59,11 @@ def _load_fbt_data(file_path: str) -> dict[str, np.ndarray]:
...
+    return eq.__dict__
```
tamaranorman, line 65: Use dict(eq) but also maybe rename to something like loaded_data or similar as eq is generally used for equals

torax/geometry.py
```diff
@@ -879,6 +880,213 @@ def from_fbt(
...
+        # --- Area, Volume, R_inboard, R_outboard
+        # --- FSA: <1/R^2>, <Bp^2>, <|grad(psi)|>, <|grad(psi)|^2>
+        # --- Toroidal plasma current
+        # --- Integral dl/Bp
+        # -----------------------------------------------------------
+
+        # Gathering area for profiles
+        areas, volumes                = [], []
+        R_inboard, R_outboard         = [], []
+        flux_surf_avg_1_over_R2_eqdsk = [] # <1/R**2>
```
tamaranorman, line 966: nit: the underscore eqdsk is probably not necessary here?

To save vertical space etc is it worth preallocating the numpy arrays and then assigning in rather than needing to create the arrays and then turn them into arrays?

tamaranorman (changes requested):
Thanks for the PR it looks really good and adds some nice functionality just a few small mainly formatting etc comments

torax/geometry_loader.py
fusionby2030: This needs a formal TODO, not sure how the torax team wants to handle COCOS in the future.

torax/geometry_loader.py
fusionby2030: dict(eq) did not work, therefore I have kept it as eq.__dict, except renaming as suggested.

torax/geometry.py
fusionby2030: I think the underscore is useful as the `from_chease` method has similar syntax for the FSA quantities.

tamaranorman:
Looks good, are you able to resolve the conflicts so that we can try it out

tamaranorman (approved):

torax/geometry.py
```diff
@@ -879,6 +880,197 @@ def from_fbt(
...
+        offset = 0.01
+        mask = (X > Xlcfs.min() - offset) & (X < Xlcfs.max() + offset) & (Z > Zlcfs.min() - offset) & (Z < Zlcfs.max() + offset)
+        masked_psi_eqdsk_2dgrid = np.ma.masked_where(~mask, psi_eqdsk_2dgrid)
+
+        # q on uniform grid (pressure, etc., also defined here)
+        q_eqdsk_1dgrid = eqfile['qpsi']
+
+        # ---- Interpolations
+        q_interp = scipy.interpolate.interp1d(psi_eqdsk_1dgrid, eqfile['qpsi'], kind='cubic')
+        psi_spline_fit   = scipy.interpolate.RectBivariateSpline(X_1D, Z_1D, psi_eqdsk_2dgrid, kx=3, ky=3, s=0)
```
jcitrin, line 932: out of curiosity, did you check robustness of the results to the hyperparameters of this method?

torax/geometry.py
```diff
@@ -879,6 +880,197 @@ def from_fbt(
...
+        F_interp = scipy.interpolate.interp1d(psi_eqdsk_1dgrid, eqfile['fpol'], kind='cubic') # toroidal field flux function
+
+        # -----------------------------------------------------------
+        # --------- Make flux surface contours ---------
+        # -----------------------------------------------------------
+
+        # set epsilon based on floating point precision of psi grid
+        # NOTE: this is hacky and should be replaced with a more robust method
+        epsilon = 1e-7
```
jcitrin, line 941: nit: can use eps from the `constants.Constants` dataclass

torax/data/third_party/geo/eqdsk_cocos02.eqdsk
jcitrin, line 1: Was this generated directly from the existing ITER_hybrid CHEASE file?

If so, and if it doesn't break the EQDSK file, would be good to mention this in a header

torax/examples/eqdsk_example.py
jcitrin, line 1: if all else apart from geometry is identical to test_iterhybrid_predictorcorrector, then can shorten this file as well as making the provenance clearer, by copying the CONFIG from iterhybrid and then modifying it. See https://github.com/google-deepmind/torax/blob/main/torax/tests/test_data/test_iterhybrid_rampup_restart.py for an example

jcitrin:
Thanks, looks really good. Made some minor comments after another run through.

After these, and merging with main, we'll bring it in internally and push it from there, next week. Thanks again for this!

torax/geometry.py
fusionby2030: IIRC, I did do some testing with this, but determined the resolution of the underlying eqdsk file is more important than the hyperparameters of the spline fitting. Perhaps a warning is of use for this.

torax/examples/eqdsk_example.py
fusionby2030: Replaced with the example. That is much cleaner.

theo-brown:
Is this PR going to be merged in any time soon? Keen to use it to smooth out the JETTO->TORAX workflow.

jcitrin:
@fusionby2030 , @theo-brown , this is now merged with [f7082a1](https://github.com/google-deepmind/torax/commit/f7082a1c503cd2fea8449606719632b9490d9712)

Some changes compared to the last PR commit:

1. Modification of minor radius definition from EQDSK, to the midplane average instead of distance from axis to LFS

2. A sim integration test was set up, with appropriate parameters for the EQDSK small-tokamak-equilibrium that was generated for the demo. @fusionby2030 , do you have further information on what this equilibrium was based on? We can add this info in a small PR. In any case it runs quite fast. The issue with the slow running sim beforehand was that the iterhybrid config was not appropriate for this EQDSK file. e.g. very high Ip compared to the EQDSK B0, leading to non-physical q<<1 , as well as extremely high power densities. This led to QLKNN to go haywire and produced chis at the max towards 100, which then led to small timesteps due to the chi time_step_calculator, which is adaptive and sets t \propto 1/chi.
  
3. Fixing minor linter issues. Recommend using pylint. 

Apologies for not being more transparent with pushing to this branch and discussing these changes. Next time when collaborating on the github version, will concentrate on carrying out the review and pushing iterative changes on github, as opposed to our internal version.

Thanks again @fusionby2030 for this very valuable feature! @theo-brown , looking forward to benchmarks against JETTO-EQDSK. Would be happy to discuss going forward.

theo-brown:
Agreed, many thanks @fusionby2030 - a major contribution to the tool!

jcitrin:
@theo-brown The CHEASE iterhybrid equivalent EQDSK landed in pint. I'm bringing it now into TORAX and will take a first look at the EQDSK-CHEASE comparison.

PR #399, opened 2024-09-12
Add EC heating and current drive

theo-brown:
Implementing the ECCD equation:

$$
j_{ec} \propto \zeta_{ec} p_{ec} \frac{T_e}{n_e}
$$

Tasks:
- [x] Set up skeleton structure
- [x] Correctly flux-surface-average EC power
- [x] Take EC power density profile as an input profile
- [x] Convert to combined heating and current drive
- [x] Modify Lin-Liu equation to use log to avoid floating point precision errors
- [x] Add option to prescribe via Gaussians
- [x] Match updated interface 
- [x] Unit tests
- [x] Docs


araju:
Hey Theo!

Sorry for the delay, but wanted to add some notes to help you add/change the source. This is a process we are in the middle of changing and documenting, but I'll leave some instructions here as well as comment your PR.

# Adding a new source

For adding a new source, you need to do the following (some of which you've already done):

## 1. Define the new source.

```python
from torax import array_typing
from torax.sources import runtime_params as runtime_params_lib
from torax.sources import source

# Define the "dynamic" runtime params that are inputs to the source.
@chex.dataclass(frozen=True)
class DynamicRuntimeParams(runtime_params_lib.DynamicRuntimeParams):
  foo_param: array_typing.ScalarFloat
  bar_param: float

# Define the runtime params class which holds the input values from the config.
@dataclasses.dataclass(kw_only=True)
class RuntimeParams(runtime_params_lib.RuntimeParams):
  foo_param: runtime_params_lib.TimeInterpolated = 1.0
  bar_param: float = 2.0

  def make_provider(self, torax_mesh: geometry.Grid1D | None = None):
    if torax_mesh is None:
      raise ValueError('torax_mesh is required for RuntimeParams.make_provider.')
    return RuntimeParamsProvider(**self.get_provider_kwargs(torax_mesh))

# Define the runtime params provider that produces "dynamic" params for each time step
# based on the values stored in the RuntimeParams.
@chex.dataclass
class RuntimeParamsProvider(runtime_params_lib.RuntimeParamsProvider):
  runtime_params_config: RuntimeParams

  def build_dynamic_params(
      self,
      t: chex.Numeric,
  ) -> DynamicRuntimeParams:
    return DynamicRuntimeParams(**self.get_dynamic_params_kwargs(t))
```

You can see examples in https://github.com/google-deepmind/torax/blob/main/torax/sources/electron_density_sources.py and https://github.com/google-deepmind/torax/blob/main/torax/sources/qei_source.py for this.

## 2. Hook up the new source to the configs.

This is the part we are actively changing. But right now, what you need to do is also modify part of TORAX to tell the config parser about your new implementation.

Add your new source to the mappings in https://github.com/google-deepmind/torax/blob/main/torax/sources/default_sources.py. For ECRH, use a new key, like `"ecrh"` and then change all the functions in that file. For the default runtime params, be sure to set the mode to `ZERO` in order to make sure you don't break any existing tests.

## 3. Start using your new source.

Once added to `default_sources.py`, the config dictionaries will be able to pick up your changes (after reinstalling depending on how you set up your repo).

```python
CONFIG = {
    'sources': {
        ...
        'ecrh': {
            'ecrh_density': 2.0,
        },
    },
}
```

# What's missing from your PR

So, from your PR right now, you need to:

 -  Add the "provider" stuff for the runtime params. The way you are doing things is a little outdated now.
 - Change `default_sources.py`

torax/sources/ec_current_source.py
```diff
@@ -0,0 +1,86 @@
...
+    # EC power density profile
+    ec_power_density: runtime_params_lib.interpolated_param.TimeInterpolated
+
+    def build_dynamic_params(self, t: chex.Numeric) -> DynamicRuntimeParams:
+        return DynamicRuntimeParams(
+            **config_args.get_init_kwargs(
+                input_config=self,
+                output_type=DynamicRuntimeParams,
+                t=t,
+            )
```
araju, line 62: As mentioned in my other longer comment, you're going to need to change this to match the style of how sources are added now. See https://github.com/google-deepmind/torax/blob/main/torax/sources/qei_source.py

araju (changes requested):

torax/sources/ec_current_source.py
theo-brown: Amazing, thanks so much for this comment. I'll get on it later this week (Thursday probably).

theo-brown:
Thanks for the step-by-step breakdown @araju! 

I have a few remaining questions:

1. Should sources still subclass based on the terms they contribute (e.g. `SingleProfilePsiSource`) or should they subclass `source.Source` directly? In `QeiSource`, it subclasses `source.Source`, despite there existing `source.IonElectronSource`, for example.

2. When should a source use `sources.source_profiles`?

3. Are my type annotations correct on the `TimeInterpolated` variables? I've attempted to follow the `jext` and `qei` conventions.

jcitrin:
A general comment which also answers Q1:

We should avoid using the intermediate "convenience classes" like SingleProfilePsiSource, SingleProfileNeSource, IonElectronSource, etc. These are unnecessary and will all soon be refactored away. Better to inherit the base Source class, and directly set `affected_core_profiles` in your new ECRH class.

Related to this, I think that this source should not only be setting the current drive. It should be a model implementation of ECRH/ECCD and set both the electron heat source and its proportional current drive.  In fact, a class `ECRHCurrentSource` or `ElectronCyclotronCurrentSource` shouldn't exist. There should just be something like `ElectronCyclotronSource` and it will have psi and temp_el `affected_core_profiles`. Then, the `runtime_params` would include options to either prescribe heating profiles directly, which are then outputted for the heating index of the output tuple,  or alternatively to set Gaussian coefficients for a Gaussian profile (and some kind of toggle to choose which prescription heating option is taken), and then in addition the current drive proportionality formula is used. The file `ec_current_source` should be called `ecrh_source` or somesuch, and then the model implementation `_calc_eccd_current` should be given a less generic name (since additional model implementations can come down the line), so something like `_prescribed_heat_chanprater_current`   . Pretty soon we'll add an additional string input to each source, which specifies which model implementation will be used.

Q2: no need to touch this. The source_profiles are automatically constructed from the config. See code in `source_models.py`. Only the sources which "must exist" (even if their values are imposed to zero based on their runtime_params), are hard-coded in source_profiles. This is the ion-electron heat exchange, and bootstrap current. These sources are always used in various parts of the code so they must exist, which also makes sense physically since these are plasma "internal sources" so to speak.

Q3: I think for the heat deposition density it should be a TimeRhoInterpolated.  Here you can follow the same pattern as `prescribed_values` in the source RuntimeParams base class https://github.com/google-deepmind/torax/blob/53390ca796bf67b2f51841e5e803301bfcacd7bd/torax/sources/runtime_params.py#L96

theo-brown:
Yes, my Q1 was just checking that I'd remembered correctly you saying this exact thing in person!

Agreed about this source setting both heating and current drive, I will make those changes today. I need to have a look at the heat source interface to make sure everything lines up.

Q2: thanks for clarifying, this makes sense!

Q3: noted, will change.

torax/sources/ec_current_source.py
```diff
@@ -0,0 +1,123 @@
...
+    #   https://doi.org/10.1088/1741-4326/ad6ea2
+
+    ne_face_in_m3 = (
+        core_profiles.ne.face_value() * dynamic_runtime_params_slice.numerics.nref
+    )
+    Te_face_in_J = core_profiles.temp_el.face_value() * CONSTANTS.keV2J
+
+    # Flux-surface averaged j profile, <j_ec> [A/m^2]
+    return (
+        CONSTANTS.epsilon0**2 # [m^-3 kg^-1 s^4 A^2]
```
jcitrin, line 99: This formula will struggle with fp32 underflows. Currently we use fp64 but reverting to fp32 is on the roadmap.

Better to do a logarithm version of the formula and then exponentiate it. See e.g. https://github.com/google-deepmind/torax/blob/53390ca796bf67b2f51841e5e803301bfcacd7bd/torax/physics.py#L81

torax/sources/electron_cyclotron_source.py
```diff
@@ -0,0 +1,178 @@
...
+@dataclasses.dataclass(kw_only=True)
+class RuntimeParams(runtime_params_lib.RuntimeParams):
+    """Runtime parameters for the electron-cyclotron source."""
+
+    # Global dimensionless current drive efficiency
+    global_efficiency: (
+        runtime_params_lib.interpolated_param.InterpolatedVarSingleAxisInput
+    )
+
+    # EC power density profile on the rho grid; units [W/m^2]
```
jcitrin, line 28: Typo: W/m^3

torax/sources/electron_cyclotron_source.py
```diff
@@ -0,0 +1,178 @@
...
+    #   Tholerus et al., 2024
+    #   https://doi.org/10.1088/1741-4326/ad6ea2
+
+    ne_face_in_m3 = (
+        core_profiles.ne.face_value() * dynamic_runtime_params_slice.numerics.nref
+    )
+    Te_face_in_J = core_profiles.temp_el.face_value() * CONSTANTS.keV2J
+
+    # Flux-surface averaged j profile, <j_ec> [A/m^2]
+    j_ec = (
```
jcitrin, line 102: should make j_ec resilient to fp32 (currently fp64 is default but that may change) . Recommend taking the exponent of the log of this function, where e.g. q**3 becomes 3*log(q)

torax/sources/electron_cyclotron_source.py
```diff
@@ -0,0 +1,178 @@
...
+class RuntimeParams(runtime_params_lib.RuntimeParams):
+    """Runtime parameters for the electron-cyclotron source."""
+
+    # Global dimensionless current drive efficiency
+    global_efficiency: (
+        runtime_params_lib.interpolated_param.InterpolatedVarSingleAxisInput
+    )
+
+    # EC power density profile on the rho grid; units [W/m^2]
+    ec_power_density: runtime_params_lib.interpolated_param.InterpolatedVarTimeRhoInput
```
jcitrin, line 29: is it OK if there is no default here? Maybe better to include zeros as default

jcitrin, line 29: May be nice to have options on how to set the power density. This one is good as a prescribed option. Can maybe change the name to: "prescribed_ec_power_density" ?

Could think of having Gaussian parameters (width, center, totalpower) which are TimeInterpolated runtime_params. Then maybe if "prescribed_ec_power_density" is None then these Gaussian ones are used, and if it isn't none then the prescribed_ec_power_density profile is used?

jcitrin: Or alternatively None as default. Depends on what choices are made vs Gaussian inputs

torax/sources/electron_cyclotron_source.py
theo-brown: Is current density in A/m^3 then, too?

theo-brown: I revisited the dimensional analysis and answered my own question!

torax/sources/electron_cyclotron_source.py
theo-brown: Great, done.

theo-brown:
Revisiting some of the maths to make sure I have the conversions correct. I think the equation I'm currently using is for the parallel current, not the toroidal current. Which does TORAX need?

jcitrin:
In the equation summary https://torax.readthedocs.io/en/latest/equation_summary.html , you can see that the non-inductive current for the psi equation is parallel to B. So parallel current is good.

theo-brown:
Ah, thanks!

For ease of searching, noting this here: the current density in the diffusion equation is
> flux-surface-averaged non-inductive current (comprised of the bootstrap current and external current drive) **projected onto and multiplied by the magnetic field**.

ie $\langle j \dot B \rangle$.

theo-brown:
Sorry this is taking so long, btw - trying to make sure that whatever I implement here is consistent with JETTO.

theo-brown:
Adding this for the full derivation: [Global electron-cyclotron current drive equation.pdf](https://github.com/user-attachments/files/17189101/Global.electron-cyclotron.current.drive.equation.pdf)

torax/sources/electron_cyclotron_source.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+
+    supported_modes: tuple[runtime_params_lib.Mode, ...] = (
+        runtime_params_lib.Mode.ZERO,
+        runtime_params_lib.Mode.MODEL_BASED,
+    )
+
+    model_func: source.SourceProfileFunction = _calc_heating_and_linliu_current
+
+    def get_value(
```
theo-brown, line 161: Can `get_value` be removed, and purely inherited from `source.Source`? This pattern was copied over from `IonElectronSource`, and I'm not sure whether that is up to date / the best way to be doing things.

torax/sources/electron_cyclotron_source.py
theo-brown: Default now is zeros.

torax/sources/tests/electron_cyclotron_source.py
theo-brown, line 1: Currently this is just a placeholder, as there appeared to be a few different versions of the test suite for sources implemented at the moment. If you can point me in the direction of the right one to copy, then that would be fab!

Nush395: hey Theo! Is the question about which of the source test cases you should inherit from here?

Assuming so, we currently have two children `SingleProfileSourceTestCase` and `IonElSourceTestCase` of `SourceTestCase` in `sources/tests/test_lib.py`. The `IonElSourceTestCase` looks most relevant for this particular source but there is one method on that test class `test_extraction_of_relevant_profile_from_output` that won't be applicable to your case so I would either:
- fork that code and put it into your own test case class.
- modify the ion el source test case to be more generally applicable to other sources that affect two core profiles.

I'd say to go for the former of forking as looking at the source test cases they could be tidied up and consolidated, so if you fork the logic in `IonElSourceTestCase` and make it applicable to your source, we will do some test consolidation clean up later as that will hopefully be the least friction for you!


If this wasn't your question ignore the above and please let me know :)

theo-brown:
Thanks @Nush395, helpful input! I've added tests for the following:
- Test it provides a value of the right shape
- Test that it modifies the right profiles
- Test that it errors if provided an unsupported mode.

However, the third test fails, and I'm struggling to debug it.
`ElectronCyclotronSource.get_value()` calls `super().get_value()`, which calls the necessary `self.check_mode` which should raise the error - except no error is raised. When debugging, the `mode` in `check_mode` is a `TracedArray` that I couldn't inspect, so I'm not quite sure what to do next.

torax/sources/electron_cyclotron_source.py
Nush395: I believe it can yes :)

torax/sources/tests/electron_cyclotron_source.py
```diff
@@ -0,0 +1,164 @@
...
+
+        core_profiles = core_profile_setters.initial_core_profiles(
+            dynamic_runtime_params_slice=dynamic_runtime_params_slice,
+            geo=geo,
+            source_models=source_models,
+        )
+
+        for unsupported_mode in self._unsupported_modes:
+            source_builder.runtime_params.mode = unsupported_mode
+            with self.subTest(unsupported_mode.name):
```
Nush395, line 118: I think you are missing a construction of a new `dynamic_runtime_params_slice` here (which will have the updated `unsupported_mode` inside it that should then throw an error 

```
source_builder.runtime_params.mode = unsupported_mode
dynamic_runtime_params_slice = (
          runtime_params_slice.DynamicRuntimeParamsSliceProvider(
              runtime_params=runtime_params,
              sources=source_models_builder.runtime_params,
              torax_mesh=geo.torax_mesh,
          )(
              t=runtime_params.numerics.t_initial,
          )
      )
with self.assertRaises(RuntimeError):
                    source.get_value(...
```

Nush395:
> Thanks @Nush395, helpful input! I've added tests for the following:
> 
> * Test it provides a value of the right shape
> * Test that it modifies the right profiles
> * Test that it errors if provided an unsupported mode.
> 
> However, the third test fails, and I'm struggling to debug it. `ElectronCyclotronSource.get_value()` calls `super().get_value()`, which calls the necessary `self.check_mode` which should raise the error - except no error is raised. When debugging, the `mode` in `check_mode` is a `TracedArray` that I couldn't inspect, so I'm not quite sure what to do next.

I think you're missing a construction of a new `dynamic_runtime_params_slice` before calling `get_value` on the source. See the corresponding test in `IonElSourceTestCase` for an example of that and have also added a comment on the test in this PR. Hope that fixes things! 🙏 

Also in general when debugging in TORAX it can be helpful to set the `TORAX_COMPILATION_ENABLED` environment variable to be `False`. See [here](https://github.com/google-deepmind/torax/blob/main/torax/jax_utils.py#L62) for where that is used. This will turn of any JIT compilation and will allow you to see the actual values of sjax.Array`s instead of those cryptic tracers 😄 .

jcitrin:
> When debugging, the mode in check_mode is a TracedArray that I couldn't inspect, so I'm not quite sure what to do next

You can try setting the TORAX_COMPILATION_ENABLED env variable to False, which may help make the variable concrete when debugging

torax/sources/tests/electron_cyclotron_source.py
theo-brown: Yep, that's the one! Thanks :)

theo-brown:
Setting ` TORAX_COMPILATION_ENABLED=False` enabled me to interactively debug to find the error, which was that I'd forgot to set `TORAX_ERRORS_ENABLED=True` - it had just been a while since I'd last written TORAX unit tests, so I forgot the obvious thing. Thanks for your help! 

@jcitrin did you see my comment about sum of Gaussians?

torax/sources/electron_cyclotron_source.py
jcitrin: Sorry I missed earlier the comment about the Gaussian option.

I think it's fine to evaluate the Gaussian at each evaluation. We already do this for the generic_ion_el_heat_source and nbi_particle_source (soon to be renamed generic_particle_source) and pellet_source. It's jitted and the Gaussian evaluation time is negligible.

A pattern which you can perhaps reuse is that of the pellet source. See https://github.com/google-deepmind/torax/blob/b3b3692b8491c9fa86d6229975afb943b641bc3c/torax/sources/electron_density_sources.py#L269 . Note that it uses `formulas.gaussian_profile` which is a reusable function. Note that this is a FORMULA_BASED source, but you can ignore that. We're going to drop FORMULA_BASED and just make it a special case of MODEL_BASED. You can use the `formula.gaussian_profile` function as part of the model_func too of course.

By a sum-of-Gaussians, do you mean that you'd like a model where different antennas will set different Gaussian depositions at different radii, and each Gaussian will need its own set of time-dependent parameters? If so, then if it's not too many Gaussians to be summed then I guess you can give each one it's own labelling in the runtime_params, and call `formulas.gaussian_profile` for each one and sum them. Another thing we could consider, which would require a bit of feature extension, is for the TimeInterpolatedInput to allow a tuple of parameters at each timeslice, and get_value time interpolates each tuple element (and this could be vectorized), and formulas.gaussian_profile is also extended for tuple input, returning the sum of Gaussians. @Nush395 what do you think of feasibility of this last point?

theo-brown: Yes, that's absolutely right - N antenna whose deposition can each be modelled as a Gaussian at a specific radius + width/height and the total EC heating profile as a sum of the Gaussian profiles.
Different machines might have different numbers of EC antennae (e.g. STEP will have something like 5-20), so I was thinking that if a Gaussian-based profile was required then it would be best to support an arbitrary number of antennae, rather than 1 or a predefined number.

theo-brown: Happy to contribute to the feature, but I wonder whether it would be better suited as a separate PR. Could do the single Gaussian profile in this PR, if that would be preferable, but if it's going to be superseded by the sum-of-Gaussians change I'm not sure how worthwhile it is.

torax/sources/default_sources.py
```diff
@@ -46,6 +48,8 @@ def get_default_runtime_params(
       return external_current_source.RuntimeParams()
     case 'nbi_particle_source':
       return electron_density_sources.NBIParticleRuntimeParams()
+    case 'echcd':
```
Nush395, line 51: hey Theo :) We've just made a change to simplify adding sources which involves removing default sources.

So you could remove these changes (or accept the file as deleted in a merge).

The only change that is required to hook up a source and make it discoverable to torax now is to register a source. Docs for this incoming soon but for letting you progress this CL will add below too. 

To register a source you can use the `register_source` method https://github.com/google-deepmind/torax/blob/main/torax/sources/register_source.py#L63.

We do this [here](https://github.com/google-deepmind/torax/blob/main/torax/sources/register_source.py#L164) for the commonly used TORAX sources but in general the idea is a user could use this method to register a source in any module (provided that module is imported by Python).

Registering a source requires you to pass in a reference to the `Source` you create and its `RuntimeParams` object and register that source under a `source_name`. The `source_name` you register under will be the thing that would identifying your source and be the key for your source if you wish to use it in the python experiment config e.g. for [here](https://github.com/google-deepmind/torax/blob/main/torax/examples/iterhybrid_rampup.py#L115) for the source registered and identified as `gas_puff_source`.

Hope this helps and sorry again or you needing to make additional changes, hopefully this will be a simpler pattern overall.

Let us know if that requires clarification or any further help!

torax/sources/electron_cyclotron_source.py
jcitrin: I think a single Gaussian profile is fine in this PR, using formulas.gaussian_profile. In a subsequent one we can extend TimeInterpolatedInput and formulas.gaussian_profile to then enable a sum of Gaussians. It's an extension rather than a superseding, so worthwhile to do it staged.

torax/sources/default_sources.py
theo-brown: No problem! New pattern looks great :) I think I've made the appropriate changes

theo-brown:
I've set the EC profile to be generated by summing a single Gaussian and a manual profile. By default, both of these are 0. 
Could instead have opted to go for setting one of them to None, but this worked more intuitively with `jnp.where`.

If the user wants a Gaussian, they can provide the location, width, and total power, and leave the manual profile as 0.
If the user wants an arbitrary profile, they can provide the full profile, and leave the Gaussian parameters as 0.

Note that this is different from what I understand the `PRESCRIBED` mode to be; to my understanding, `PRESCRIBED` should set the output of a source directly. In this case, that would mean setting a power density profile (for the `TEMP_EL` equation) **and** a `<j.B>` current density profile (for the `PSI` equation). Instead, here we are providing two options for setting an EC power density, and the `<j.B>` profile is computed according to the Lin-Liu/Prater model. Hence, it is still model-based, rather than prescribed.

jcitrin:
Great, thanks. I guess test_bohmgyrobohm_all.nc needs to be regenerated?

theo-brown:
Is that comment for the other PR https://github.com/google-deepmind/torax/pull/447?

jcitrin:
Yes it was, thanks. Got my wires crossed

jcitrin:
Yes, the Gaussian implementation looks good. Can proceed with the docs and we'll do a review, will be just some minor comments.

theo-brown:
As the cited reference is in terms of j_tor, rather than <j.B>, I wonder whether I should also add the following derivation to the docs somewhere.
[Electron-cyclotron current drive efficiency in terms of j.B.pdf](https://github.com/user-attachments/files/17434437/Electron-cyclotron.current.drive.efficiency.in.terms.of.j.B.pdf)

theo-brown:
Yes, it's 4 orders of magnitude smaller, even in ST geometry. Some example plots from a STEP case:

image

image

jcitrin:
Great thanks. I just tried with other settings (high current, low area) and got 2-3 orders of magnitude lower. Will mention O(1e-3-1e-5) in the docs

jcitrin:
@theo-brown this has just been merged with [357f532](https://github.com/google-deepmind/torax/commit/357f53234adcf2aef8b4b317796e406bc0738221)

Additions here compared to the last PR commit:

1. Various linter fixes (recommend using pylint to reduce these)
2. Bugfix in current drive formula (there was a * instead of + in the log formula on the epsilon0 line)
3. Set default mode=MODEL_BASED
4. Split electron_cyclotron_source output to _el (electron heat) and _j (current) profiles in the xr dataset
5. Added the new profiles to the plotting routines, and fixed some minor bugs found in pre-existing plotting routine code (in legend labels)
6. Made a new integration sim test based on iterhybrid_predictor_corrector, replacing generic_current with the new ECCD model, and removing the electron heating from generic_ion_el_heat_source.
7. Added total ECRH current and total generic_current to post_processing data and plotting routines. Helped with the sanity checking and also good outputs to have.
8. Reran all test cases to propagate new outputs
9. Fixed psidot calculation to include sources like the new one, that have both a PSI AffectedCoreProfile as well as other AffectedCoreProfiles.

theo-brown:
Excellent changes, thank you so much! 

Good catch on the typo in (2).

I'll try and fix my linter settings - I currently use Ruff (because it's very fast and does auto-formatting), I'd tried to export the .pylintrc file into a ruff.toml but I think the conversion must've been lossy. I'll switch over to pylint for future.

Issue #403, opened 2024-09-16
[bug] `NaN` in `d_face_el` and `v_face_el` when using Bohm-gyroBohm


theo-brown:
**Behaviour**: simulation exits due to `nan` in core profiles when run with the BgB transport model.
**Diagnosis**: In some scenarios, `chi` terms are zero on axis. This causes the `D_face` on axis to be `nan` (as it divides by `chi`), and hence the `V_face` to be `nan` (as it is proportional to `D_face`).

theo-brown:
A simple solution would be to add a small term to the `chi`s. I have done this manually and the simulations run; however, I'm going to do some investigating to see how other codes (RAPTOR, JETTO) handle the same problem.

jcitrin:
Yes, please compare to RAPTOR and JETTO and we can decide on the fix. 

But for D it seems pretty clear that the zero-gradient boundary condition means D=0 on-axis: chi_e=chi_i=0 on axis (see eq 8+9 in Tholerus 2024), and thus in eq.10, we have 0^2/0, i.e. 0. So can define  `d_face_el` as something like:

```
d_face_el = jnp.concatenate((jnp.zeros(1), weighting[1:] * chi_e[1:] * chi_i[1:] / (chi_e[1:] + chi_i[1:]))))
```

Same approach for V. It scales with S^2/(Volume*Volume') , and since (roughly) S= pi r^2 and Volume = 2pi^2r^2R , V scales with r and is zero on axis.

NOTE: there seems to be a bug in `v_face_el`. It's missing a `geo.rho_b` term, when comparing to equation 12 in Tholerus 2024. Can be fixed in the same opportunity as fixing the divs by zero.

jcitrin:
If you'd like to send the PR go ahead, otherwise can do it on our side. Would be good to add a BgB case with particle transport to the test suite.

theo-brown:
RAPTOR adds its inner transport patch, a neoclassical term, and a sawtooth term before `D` and `v` are calculated ([line 276 in `chi_BgB.m`](https://gitlab.epfl.ch/spc/raptor/raptor/-/blob/main/code/physics/chi_BgB.m?ref_type=heads#L101)).

I'm not proficient in FORTRAN, but from my initial scan of the JETTO repo I think it sets `D=0` where `chi=0` ([line 224 in `callcgm.f`](https://git.ccfe.ac.uk/jintrac/jetto-sanco/-/blob/master/jetto/callcgm.f#L224)).

[...]

PR #406, opened 2024-09-17
Fix Bohm-gyro-Bohm model particle transport

theo-brown:
Implement fixes described in #403.

**Tasks:**
- [x] Add missing `rho_b` term to `v_face_el`
- [x] Solve `nan` on-axis for `D_face_el`
- [x] Add particle transport test case


jcitrin:
For the test case: to avoid test bloat perhaps it's better to just extend the existing BgB test to include particle transport as well

theo-brown:
Note that my plasma experience is not strong enough yet to sense-check the results of the test case.

jcitrin:
Thanks for all this.

About sense-checking the results. Here's what it looks like:

image


The bulk ne profile looks OK, but things are a bit wonky in the "pedestal" region. This is because of inappropriate defaults for this case. The ne boundary condition default is nebound=0.85*nGW (0.85 Greenwald fraction), but neped is set as 0.7e20, which ends up under that value for this case. We later plan to introduce machine-specific defaults for config, but are not there yet.

When introducing more appropriate density settings, i.e.
```
    'runtime_params': {
        'profile_conditions': {
            'nbar': 0.8,
            'neped': 0.8,
            'neped_is_fGW': True,
            'ne_bound_right': 0.5,
            'ne_bound_right_is_fGW': True,
        },
```

we get 
image

which looks better.

Only other comment for the review is that the name BgB test case in tests/sim.py also needs to be updated.

I've done above modifications on our end, and this will included in the merge, so no further action is necessary :)

jcitrin:
Merged with ab28a713

PR #572, opened 2024-11-26
Add simple impurity radiation model (P_imp as a fraction of P_tot_in)

theo-brown:
Add a simplified radiation sink model, `P_rad = α*(P_aux + P_alpha + P_ohmic)`

This is based on the functionality in JETTO:
image

Todo:
- [x] Calculate prad as a fraction of ptot
- [x] Add to outputs and plotting
- [x] Unit tests
- [x] Integration tests


theo-brown:
@Nush395 if you've got time, could you lend a hand with this? I have a couple of questions about implicit sources and sources that link back.

At the moment I'm finding that my implementation is nonzero when I inspect it during runtime but has no effect on the state. In the output, it is 0.

Nush395:
> @Nush395 if you've got time, could you lend a hand with this? I have a couple of questions about implicit sources and sources that link back.
> 
> At the moment I'm finding that my implementation is nonzero when I inspect it during runtime but has no effect on the state. In the output, it is 0.

Yes for sure! Happy to answer questions, feel free to drop them on this PR or we can go over on a call if that's easier :)

Nush395:
> @Nush395 if you've got time, could you lend a hand with this? I have a couple of questions about implicit sources and sources that link back.
> 
> At the moment I'm finding that my implementation is nonzero when I inspect it during runtime but has no effect on the state. In the output, it is 0.

Just took a quick look at the change. Just a sanity check, have you run the `prad` config with 
```
'radiation_heat_sink': {'mode': 'model_based'}
```
?
(As the default mode that you inherit from the `RuntimeParams` is set to ZERO.

theo-brown:
Ah, thanks for spotting that. That would be why it wasn't being used! I'll do that and then see where I get to. Sorry for not spotting the obvious thing!

Nush395:
> Ah, thanks for spotting that. That would be why it wasn't being used! I'll do that and then see where I get to. Sorry for not spotting the obvious thing!

No worries at all, I've run into that gotcha myself!

theo-brown:
Note: as of 69f5ada, unit tests fail because I've added a term to post_processed_outputs that is not present in all of the current `.nc` files. Hence, all test data will need to be regenerated.

Nush395:
> As of [69f5ada](https://github.com/google-deepmind/torax/commit/69f5ada5c269ef3f67ea285a879cf0ae0a83dda5), unit tests fail because I've added a term to post_processed_outputs that is not present in all of the current `.nc` files. Hence, all test data will need to be regenerated. Is there a speedy way of doing this?

There is, there is a `run_and_save_all_benchmarks.py` script that we use. See the bottom of https://torax.readthedocs.io/en/latest/contribution_tips.html#contribution-tips that @jcitrin recently added to. (Will also make this a bit more visible, if you have suggestions on where this should be please let us know!)

theo-brown:
For the time being I've gone for only inheriting from `SourceTestCase`, which means:
- Inherit
  - `test_expected_mesh_states`
  - `test_runtime_params_builds_dynamic_params`
- Reimplement
  - `test_extraction_of_relevant_profile_from_output`
  - `test_invalid_source_types_raise_errors`
  - `test_source_value`

I couldn't quite figure out why `test_extraction_of_relevant_profile_from_output` and `test_invalid_source_types_raise_errors` weren't implemented in the parent class - it seems to me that they should be functions of the source properties only? Regardless, this seemed to be the pattern implemented by most other sources, so I went with that.

Nush395:
> I couldn't quite figure out why `test_extraction_of_relevant_profile_from_output` and `test_invalid_source_types_raise_errors` weren't implemented in the parent class - it seems to me that they should be functions of the source properties only? Regardless, this seemed to be the pattern implemented by most other sources, so I went with that.

For `test_invalid_source_types_raise_errors` that is implemented in the `SingleProfileSourceTestCase` and you are right for the `test_extraction_of_relevant_profile_from_output` this is likely a legacy thing that could be tidied up and pushed into the parent class now. I can TAL at doing that and doing some test tidying up.

theo-brown:
>  I can TAL at doing that and doing some test tidying up.

I'm more than happy to tackle this in a separate PR if you want to have a look at the ohmic source? Unless you'd rather do both!

Nush395:
> > I can TAL at doing that and doing some test tidying up.
> 
> I'm more than happy to tackle this in a separate PR if you want to have a look at the ohmic source? Unless you'd rather do both!

That would be great if you have time, thanks!

torax/tests/test_data/test_changing_config_before_LOCAL_171788.nc
jcitrin, line 1: can remove

torax/tests/test_data/test_changing_config_before_REMOTE_171788.nc
jcitrin, line 1: can remove

torax/tests/test_data/test_implicit.nc
jcitrin, line 1: how come this needed to be replaced?

torax/sources/radiation_heat_sink.py
```diff
@@ -0,0 +1,124 @@
...
+) -> jax.Array:
+    """Model function for radiation heat sink.
+
+    In this model, a fixed % of the total power input to the temp_el equation is lost."""
+    # Based on source_models.sum_sources_temp_el and source_models.calc_and_sum_sources_psi,
+    # but only summing over heating *input* sources (Pohm + Paux + Palpha)
+    # and summing over *both* ion and electron heating
+
+    def get_temp_el_profile(source_name: str, source: source_lib.Source) -> jax.Array:
+        # TODO: Currently this recomputes the profile for each source, which is inefficient
```
jcitrin, line 38: agree. We will prioritize changing the pattern such that this won't happen.

torax/examples/prad.py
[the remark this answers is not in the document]
theo-brown: Yes, I'll definitely do this. I hadn't actually meant to leave this config in, I'd forgotten that I hadn't got round to turning it into an integration test yet!

torax/sources/radiation_heat_sink.py
theo-brown: OK for the time being though?

theo-brown:
Regarding the changed .nc files: these are due to `torax/tests/test_run_simulation_main.py`. 

From what I can tell, tests in this file produce output .nc with all of the post-processed outputs, regardless of whether the corresponding source etc are enabled. I couldn't quite pinpoint why this is the case. It's most likely a bug.

As I've added `P_imp` to the post-processed outputs, I needed to regenerate these files.

jcitrin:
>  From what I can tell, tests in this file produce output .nc with all of the post-processed outputs, regardless of whether the corresponding source etc are enabled. I couldn't quite pinpoint why this is the case. It's most likely a bug.

This is a feature not a bug :)  . Having zero integrated powers written even when the source isn't included in the simulation was part of the design. The idea was more robust analysis routines and avoiding needing to handle non-existing attributes. However this is not the case for the non-integrated source outputs, so it's a bit inconsistent. We will consider revisiting this in the near future as part of an ongoing output overhaul, and make a single convention.

torax/tests/test_data/test_iterhybrid_predictor_corrector_impurity_radiation.py
```diff
@@ -0,0 +1,23 @@
...
+
+"""Identical to test_iterhybrid_predictor_corrector but simple impurity radiation."""
+
+import copy
+
+from torax.tests.test_data import test_iterhybrid_predictor_corrector
+
+CONFIG = copy.deepcopy(test_iterhybrid_predictor_corrector.CONFIG)
+
+CONFIG['sources']['impurity_radiation_heat_sink'] = {}
```
jcitrin, line 23: nit: while there is a default, in practice users will almost always set the specific fraction. So maybe good to demonstrate that with adding the runtime_param here.

torax/plotting/configs/default_plot_config.py
```diff
@@ -145,8 +145,8 @@
             suppress_zero_values=True,  # Do not plot all-zero data
         ),
         plotruns_lib.PlotProperties(
-            attrs=('q_brems', 'q_rad'),
-            labels=(r'$Q_\mathrm{brems}$', r'$Q_\mathrm{rad}$'),
+            attrs=('q_brems', 'q_imp'),
```
jcitrin, line 148: nit: sorry for being so nitpicky, but while the source name is impurity_radiation_heat_sink , I think it's OK to maintain the shorthand q_rad, P_rad, Q_rad, which are all user-facing and standard convention.

We'll then have say q_rad, q_brems, q_cycl (coming soon) and this will be clear enough for fusion physicists

jcitrin (approved):
Looks good. A couple of small nits. Will bring it in on our end once they're done.

torax/plotting/configs/default_plot_config.py
theo-brown: Sure, no problem.

jcitrin:
Great thanks. Bringing it in. That failing test is due to the next JAX release, see #591 . No worries, should be fine in the final merge

jcitrin:
Merged by 0d66813

PR #658, opened 2025-01-15
Add ne_greenwald_fraction to PostProcessedOutputs

theo-brown:
The Greenwald density fraction is generally useful. Having it in the simulation state is useful as I am interested in developing feedback controllers for density fuelling, with the Greenwald fraction as the target variable (eg #532).


jcitrin:
Thanks for this.

Can have a short discussion on appropriateness of line-averaged-density vs volume-averaged-density.

In the literature I think line-averaged-density is more often used. For a given machine it's easier to measure (no need for full geometry reconstruction) but at the same time the specific chord used is very machine specific.

Volume-averaged-density is less ambiguous.

However, if you are developing controllers for density feedback, then should think about what the appropriate synthetic diagnostic is. Is the fGW provided by the real diagnostic a line average with a specific chord, or volume averaged following geometry reconstruction? Likely the former, and then you'd probably want a more realistic synthetic diagnostic in TORAX with the actual chord, that would then necessitate more 2D geometry information propagated in TORAX etc etc.

For now happy to approve this volume averaged one, but wanted to raise the above points. Feel free to include some of these points in the docstring of the method if relevant. Am also wondering if the name should change to e.g. ne_volume_avg_fgw, or fgw_ne_volume_avg, or something like that, to accentuate the density used for the fraction.

theo-brown:
> you'd probably want a more realistic synthetic diagnostic in TORAX with the actual chord, that would then necessitate more 2D geometry information propagated in TORAX etc etc

Very interested in this longer term! At the moment in JETTO we tend to use density feedback on the volume averaged FGW just to make things a bit easier as the operating points are super unstable. This avoids any synthetic diagnostic paraphernalia, but is indeed less realistic from a control perspective. 

> Feel free to include some of these points in the docstring of the method if relevant. Am also wondering if the name should change to e.g. ne_volume_avg_fgw, or fgw_ne_volume_avg, or something like that, to accentuate the density used for the fraction.

Good points. In IMAS I believe[ global_quantities/greenwald_fraction](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/summary.html#summary-global_quantities-greenwald_fraction) is indeed set from the line average, so I'll change the naming for this variable to reflect the volume-average.

jcitrin:
> In IMAS I believe[ global_quantities/greenwald_fraction](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/summary.html#summary-global_quantities-greenwald_fraction) is indeed set from the line average, so I'll change the naming for this variable to reflect the volume-average.

OK great, will bring it in after the renaming

jcitrin:
Also dealing with the merge conflict and bringing it in now

jcitrin:
Some tests fail due to me not regenerating the subset of required nc files after "resolving" their merge conflict. Am regenerating all nc tests anyway on our side (even beyond those "required") such that the new output will be in all test nc files in the repo. So the final merge commit will pass the tests.

theo-brown:
Sorry for a) the typo and b) not fixing the merge conflicts myself! Thanks for bringing it in.

jcitrin:
Merged with af447f23e6c483c472d30d88122729b0d2b90b0d

Changes in the final submitted commit (from internal repo) compared to last commit in this PR:
1. Using geo.Rmin instead of geo.rmid[-1] for Greenwald density limit
2. Removed superfluous comments (which repeated content of function names)
3. Added a light unit test

PR #671, opened 2025-01-22
Set BgB electron convectivity proportional to diffusivity

theo-brown:
Simulations with the BgB model were showing poor agreement in the shape of the convectivity profiles with similar simulations in other codes.

While JETTO/JINTRAC has quite a complex approach for convectivity (that I have struggled to parse from the source code*), RAPTOR adopts a much more simple method: setting `v_e ∝D_e`. In this PR I've switched the TORAX BgB model to use the same method as RAPTOR.

\* What I had implemented previously in the BgB model was only an *additional* particle pinch term, rather than the main particle convectivity. This became apparent when comparing the profiles.


docs/physics_models.rst
```diff
@@ -213,7 +212,9 @@ TORAX currently offers three transport models:
 
     :math:`c_2 = 0.3`
 
-    We note that the Bohm-GyroBohm model TORAX implementation is presently
+    :math:`c_v = 0.1`
```
jcitrin, line 215: This is an outward pinch, right? As in with no source, we'll end up with a hollow density profile (negative R/Ln)?

If so recommend making a small inward pinch the default. The magnitude seems fine, just wondering about the sign.

jcitrin (approved):

docs/physics_models.rst
theo-brown: Correct, fixed in [a9f85fc](https://github.com/google-deepmind/torax/pull/671/commits/a9f85fce64fd99e2aab282cea991150dacc79e8a)

jcitrin:
Thanks. I'll quickly fix the merge conflict (will need to regenerate the test .nc file) and bring this in now

jcitrin:
Merged with [ec8513d](https://github.com/google-deepmind/torax/commit/ec8513d94edb484d13ef67dc1ea00acbf49df694)

PR #816, opened 2025-03-10
Add validation check to prevent bremsstrahlung and Mavrin models from…

demoncoder-crypto:
… being active simultaneously

### DIFF
--- a/torax/sources/source_models.py
+++ b/torax/sources/source_models.py
@@ -319,6 +319,29 @@ class SourceModelsBuilder:
     self.source_builders = source_builders
 
   def __call__(self) -> SourceModels:
+    # Check if bremsstrahlung and Mavrin models are both active
+    bremsstrahlung_name = 'bremsstrahlung_heat_sink'
+    impurity_radiation_name = 'impurity_radiation_heat_sink'
+    
+    if (bremsstrahlung_name in self.source_builders and 
+        impurity_radiation_name in self.source_builders):
+        
+        bremsstrahlung_params = self.source_builders[bremsstrahlung_name].runtime_params
+        impurity_params = self.source_builders[impurity_radiation_name].runtime_params
+        
+        # Check if both sources are active (not in ZERO mode)
+        bremsstrahlung_active = bremsstrahlung_params.mode != runtime_params_lib.Mode.ZERO
+        
+        # Check if impurity radiation is active and using the Mavrin model
+        impurity_active = impurity_params.mode != runtime_params_lib.Mode.ZERO
+        using_mavrin = getattr(impurity_params, 'model_func', 'impurity_radiation_mavrin_fit') == 'impurity_radiation_mavrin_fit'
+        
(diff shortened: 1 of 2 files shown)


torax/sources/source_models.py
```diff
@@ -319,6 +319,29 @@ def __init__(
...
+        
+        # Check if impurity radiation is active and using the Mavrin model
+        impurity_active = impurity_params.mode != runtime_params_lib.Mode.ZERO
+        using_mavrin = getattr(impurity_params, 'model_func', 'impurity_radiation_mavrin_fit') == 'impurity_radiation_mavrin_fit'
+        
+        if bremsstrahlung_active and impurity_active and using_mavrin:
+            raise ValueError(
+                f"Both {bremsstrahlung_name} and {impurity_radiation_name} with Mavrin model "
+                f"should not be active at the same time to avoid double-counting radiation losses. "
+                f"Please set one of them to Mode.ZERO."
```
jcitrin, line 343: set one of them to Mode.ZERO or remove on of these sources from the config entirely.

jcitrin (commented):
I think this logic would work, but we are currently in the midst of a major config feature upgrade with pydantic, and all config validations should be with pydantic.

I think that much of this logic could/should be moved to be a pedantic after validator on the sources config. https://github.com/google-deepmind/torax/blob/main/torax/sources/pydantic_model.py

@Nush395 please advice on how this should be done.

jcitrin:
And thank you for your contribution! Please sign the CLA otherwise merging any PRs would not be possible

demoncoder-crypto:
I have signed the Cla i dont know why it isnt updating

torax/sources/source_models.py
demoncoder-crypto: I will work on this Asap, if validation approach should use Pydantic validators instead of direct checks, I will work on that. if its needed. let me know how to proceed. Thanks

jcitrin:
> I have signed the Cla i dont know why it isnt updating

I checked and while indeed your GitHub username passes the CLA, your email address does not. Please see the information here and follow any relevant instructions. https://github.com/google-deepmind/torax/pull/816/checks?check_run_id=38534290069 . In general, I believe you should have signed the CLA with the same email address as used for the git commits. You can ensure that the correct email address is used in the git config. You can also retroactively change the email address is the commits and/or squash your commits (rebase) after changing the email address in your git config to the correct one

tamaranorman:
So the pydantic changes that would support this are currently out for review https://github.com/google-deepmind/torax/pull/826 and should be in very soon

Once in this should go in as a check on the pydantic model torax/sources/pydantic_model.py as part of the validation there

jcitrin:
Closing this since it should be redone using pydantic model_validators. Better to start from scratch.

demoncoder-crypto:
Yes I understand. I am working on it and starting from scratch for both the issues. I need maybe a day for both of them

demoncoder-crypto:
I have already started on the other one since yesterday. Will fix the imports on it later

PR #823, opened 2025-03-11
Akash4467/fix-673

Akash4467:
Added a switch between L- and H- modes based on the plasma state


Akash4467:
@jcitrin Could you please review my PR? Thanks!

jcitrin:
The basic required behaviour is to switch, at the beginning of the computational step, the set_pedestal attribute from False→True and/or True→False depending on the ratio of Psol and Plh.  However, this PR unfortunately does not make sense for numerous reasons.

1. Defines a function that is never used
2. Uses attributes of dataclasses that do not exist
3. Compares variables of different dimensions (e.g. comparing current and power)
4. The logic (even assuming that nonexisting attributes exist) doesn't carry out the required behaviour.
5. No testing of required behaviour

I am closing this PR. I recommend not sending PRs without a clear design related to the actual requirements, and testing. It may be better if this issue is tackled by a domain expert, since there are also additional fusion physics subtleties to consider. In addition, it's better to modify the pedestal_model API , e.g. by moving set_pedestal there and including various options for the trigger model. I will update the issue description with some of the above. Thank you for triggering some deeper consideration about the requirements for this issue.

Issue #672, opened 2025-01-22
Add an absorbed power fraction to the generic heat source description


anteplu:
- the input (injected) power is prescribed in the config;
- an absorbed power fraction has to be added to the config;
- the generic heat source module has to be updated such that the heat transport equations use the absorbed power rather than injected one;
- in the output both injected and absorbed powers are somehow available;
- the fusion gain Q in the output is computed using the injected power.

demoncoder-crypto:
Hey, i would love to work on this issue, can you assign me?

demoncoder-crypto:
Hey, i tried to solve the problem. Let me know if its up to expectation or anything needs to be changed.

jcitrin:
> [@jcitrin](https://github.com/jcitrin) [@tamaranorman](https://github.com/tamaranorman) could you please take a look 

@anteplu yes we are iterating extensively on the PR

jcitrin:
Closed by #884

PR #831, opened 2025-03-12
Add an absorbed power fraction to the generic heat source description #672

demoncoder-crypto:
sync to head and resolve the merge conflicts.


demoncoder-crypto:
I have updated CLA now this is redundant

PR #834, opened 2025-03-13
Speeding up the plotting slider responsiveness 

### LINKED ISSUE #553: Speed up plotting slider responsiveness

harshtech123:
/closes #553 

this pr is indeed for increasing the responsiveness of plotting time slider.

during testing i found that the slider itself is not an issue but the computation was , to resolve this i adopted following approach -
1 use the down sample on initial slider event (low resolution but not that low)
2 increasing slider responsiveness (asserted by consoling the time taken) 
3 when the slider stops updating that down sample to full resolution 
4 i see that whenever we fires two events one after another both are getting processed (since python is single threaded on main gui) but what if we fires many events ?

- current approach handle this as handling each event 
- in this implementation i try to cancel that extra events it works when _update is fast enough and/or is broken into smaller chunks (here we check between each attribute update)


harshtech123:
current vs improved responsive ness

Screenshot 2025-03-12 224401
Screenshot 2025-03-13 151245
Screenshot 2025-03-13 143501

harshtech123:
@jcitrin  Could you please review this PR? Thanks! 🙌

jcitrin:
Thanks for your efforts. However not going to merge this. 

Not all the changes are sufficiently motivated and streamlined, and when testing on my system there are two issues:
1. Sliding response is not faster (that's the main requirement)
2. Plotting has some "flickering" of the plots when sliding, leading to a worse user experience.

harshtech123:
yes @jcitrin  i feels the same , i know this pr needs more improvement , wdyt  about how we can resolve this !
i  also try to use some other methods like debouncing to resolve this , i think we should think about more interective liberary or future optimize the calculations? thank you !

jcitrin:
If you'd like to try again then go ahead. But please ensure that there is an actual improvement in plotting behavior before reopening the PR

PR #841, opened 2025-03-14
added the extra optional fields (chi_e_bohm, chi_e_gyrobohm, chi_i_bohm, chi_i_gyrobohm) to explore raw bohm and raw gyrobohm

### LINKED ISSUE #548: Explore Bohm and GyroBohm terms separately in output

harshtech123:
/closes #548 

this pr is indeed for exploring each bohm and gyrobohm  separately in output .
this give us flexibility for diagnostics and parameter Studies: Checking how sensitive your model is.

@jcitrin can you please review and let me know if any change is required , thank you !


torax/transport_model/bohm_gyrobohm.py
```diff
@@ -106,7 +106,15 @@ def sanity_check(self):
...
 class BohmGyroBohmTransportModel(transport_model.TransportModel):
-  """Calculates various coefficients related to particle transport according to the Bohm + gyro-Bohm Model."""
+  """Calculates various coefficients related to particle transport according to the Bohm + gyro-Bohm Model.
+
+  In this modified version, the returned CoreTransport structure includes the individual
+  contributions for the Bohm and GyroBohm terms as additional fields:
+    - chi_e_bohm: Bohm contribution for electron heat conductivity.
+    - chi_e_gyrobohm: GyroBohm contribution for electron heat conductivity.
+    - chi_i_bohm: Bohm contribution for ion heat conductivity.
+    - chi_i_gyrobohm: GyroBohm contribution for ion heat conductivity.
```
jcitrin, line 116: don't modify the docstring.

"In this modified version" does not make sense to have.

torax/transport_model/bohm_gyrobohm.py
```diff
@@ -137,7 +145,9 @@ def _call_implementation(
       pedestal_model_outputs: Output of the pedestal model.
 
     Returns:
-      coeffs: The transport coefficients
+      coeffs: The transport coefficients encapsulated in a CoreTransport structure.
+             Now includes the individual Bohm and GyroBohm contributions as:
+             - chi_e_bohm, chi_e_gyrobohm, chi_i_bohm, and chi_i_gyrobohm.
```
jcitrin, line 150: don't modify the docstring.

"Now includes" does not make sense to have.

torax/transport_model/tests/bohm_gyrobohm_test.py
jcitrin, line 1: can remove this all now, and ensure that the other PR with the bohm_gyrobohm_test is merged first, then build on top of that

jcitrin (commented):
In addition to the individual comments, a couple of general things:

1. When the bohmgyrobohm model is not used, the extra optional attributes to CoreTransport will lead to a bunch of "useless" attributes with all zeros that will also be smoothed in `transport_model._smooth_coeffs`. To avoid unnecessary work, need to ensure that smoothing is only done for non-all-zero attributes

2. In `_save_core_transport` in `output.py`, need to also save to xarray the new bohmgyrobohm attributes in the case that there is any non-zero value in them (I.e. bohmgyrobohm is being used)

harshtech123:
thank you for this review we will do the same as  discussed once we finish this https://github.com/google-deepmind/torax/pull/848

jcitrin:
#848 now merged. If you like can sync this branch and continue

torax/transport_model/bohm_gyrobohm.py
harshtech123: removed

harshtech123:
@jcitrin made changes as per our discussion also updated `transport_model._smooth_coeffs` to avoid unnecessary work and  ensuring that smoothing is only done for non-all-zero attributes.
plus `_save_core_transport ` is not saving to xarray the new bohmgyrobohm attributes in the case that there is any non-zero value in them (I.e. bohmgyrobohm is being used)

thank you !

jcitrin (approved):
Thanks this looks good.

jcitrin:
Made a couple of extra commits following closer inspection on my end. Sorry I didn't catch these things in the review. Will bring this in now and merge.  Thanks again for your contributions.

1. The new transport_model post_init was leading to TORAX crashes. Unfroze the dataclass (not really necessary to freeze) and simplified the post_init.
2. There wasn't a test on whether the new bohm + gyrobohm outputs sum up to the total expected fluxes. Made the test and then found a bug in the new bohm_gyrobohm logic which I fixed.

Regarding (1), I strongly recommend you run the tests also on your end to see if they pass, before sending the PR or asking for review.

harshtech123:
@jcitrin yes i tested the test on my local , they passed ( but i only run the bohm_gyrobohm tests maybe that was the fault ) i'll keep this in mind , thank you for this review!

one more thing that i need to get cleared , i am getting file not found errors ( related to qlknn_transport_model ) when running pytest i am inspecting it from my side (since this is not comming before maybe you might give me some hint) , apologize for above  missing i realised that i'll added that in before commit but missed when resolving conflict

jcitrin:
> one more thing that i need to get cleared , i am getting file not found errors ( related to qlknn_transport_model ) 

Did you follow the instructions for installing QLKNN10D in the installation guide?

https://torax.readthedocs.io/en/latest/installation.html

harshtech123:
thank you so much ! works fine for me now .

PR #895, opened 2025-03-24
Check validity of EQDSK contour routine

theo-brown:
Prompted by #809.

Checks added:
- Ensure flux surfaces are increasing in volume

### DIFF
--- a/torax/geometry/standard_geometry.py
+++ b/torax/geometry/standard_geometry.py
@@ -866,6 +866,18 @@ class StandardGeometryIntermediates:
         / (F_eqdsk * flux_surf_avg_1_over_R2_eqdsk)
     )
 
+    # Sense-check the profiles
+    dvolumes = np.diff(volumes)
+    if not np.all(dvolumes > 0):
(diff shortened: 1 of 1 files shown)


torax/geometry/standard_geometry.py
```diff
@@ -866,6 +866,18 @@ def calculate_area(x, z):
         / (F_eqdsk * flux_surf_avg_1_over_R2_eqdsk)
     )
 
+    # Sense-check the profiles
+    dvolumes = np.diff(volumes)
+    if not np.all(dvolumes > 0):
+      idx = np.nonzero(dvolumes <= 0)
```
jcitrin, line 872: nit: should use np.where here instead

theo-brown: For readability? They're functionally the same if you're not using the 'select' property of where.

theo-brown: Makes sense! Just wanted to make sure I understood the reasoning, ty!

jcitrin (approved):

Issue #543, opened 2024-11-19
Further JAX performance optimization


jcitrin:
Better vectorization in calc_coeffs?

jcitrin:
Preferred internal development. Assignees now reflect this.

jcitrin:
No need for a separate issue, this is continuously ongoing

PR #908, opened 2025-03-26
Better vectorization in calc_coeffs 

harshtech123:
related to #543 

hii @jcitrin i try to refactor cal_coeffs in manner that it will vectorize in more better format 

here are some changes that i try to implement in this pr

-  reduce overhead from multiple elementwise operations
-  extracting the common factor into a single variable and then use it in both expressions (`full_chi_face_ion ` and `full_chi_face_el `)
-  i found out use_pereverzev a static parameter, not a dynamic one. Consequently, we can treat it as a compile-time constant and pass it as a static argument to @functools.partial

also i found that we have parameters that are common (shared across all parallel computations) and which might be batched (varying per state) `theta_imp` `use_pereverzev` `torax_mesh` `transport_model` `explicit_source_profiles` `source_models`  `pedestal_model ` but some are dynamic like `dynamic_runtime_params_slice ` and `core_profiles`

needs some confirmation before making any change regarding this  , thank you so much!

### DIFF
--- a/torax/fvm/calc_coeffs.py
+++ b/torax/fvm/calc_coeffs.py
@@ -168,25 +168,20 @@ def _calculate_pereverzev_flux(
       * geo_factor
   )
 
-  # remove Pereverzev flux from boundary region if pedestal model is on
-  # (for PDE stability)
-  chi_face_per_ion = jnp.where(
-      jnp.logical_and(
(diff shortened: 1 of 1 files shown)


jcitrin:
We have some more extensive work planned for this topic on our end. Am closing the PR and assigning, to reflect this. Apologies for that.

PR #1233, opened 2025-05-22
Add CombinedTransportModel

### LINKED ISSUE #550: Combined/additive transport from multiple transport models in different radial ranges

theo-brown:
- [x] Add Combined model with _call_implementation
- [x] Add min/max rho parameter for TransportModelBase
- [x] Raise error if patches are set in children
- [x] Unit tests
- [x] Integration tests
- [x] Docs

Closes #550


torax/_src/transport_model/combined.py
```diff
@@ -0,0 +1,103 @@
...
+        dynamic_runtime_params_slice.transport.transport_model_params,
+    ):
+      # Each transport model expects dynamic_runtime_params_slice.transport to
+      # be its corresponding DynamicRuntimeParams
+      # TODO: This seems like a hacky way of getting round what is potentially
+      # a problem with the design [TB 22 May 2025]
+      component_dynamic_runtime_params_slice = dataclasses.replace(
+          dynamic_runtime_params_slice,
+          transport=component_params,
+      )
```
theo-brown, line 70: @tamaranorman thoughts on this? Bit clunky, but I didn't know how far to unravel.

Example: `critical_gradient.CriticalGradientModel` expects `dynamic_runtime_params_slice.transport` to be of type `critical_gradient.DynamicRuntimeParams`.

https://github.com/google-deepmind/torax/blob/3c8c7764d31369651ed42edec8cd2f0d4608dca2/torax/_src/transport_model/critical_gradient.py#L79C1-L81C6

torax/_src/transport_model/runtime_params.py
```diff
@@ -47,10 +49,3 @@ class DynamicRuntimeParams:
   rho_outer: array_typing.ScalarFloat
   smoothing_width: float
   smooth_everywhere: bool
-
-  def __post_init__(self):
-    jax_utils.error_if(
-        self.rho_outer,
-        self.rho_outer <= self.rho_inner,
-        'rho_outer must be greater than rho_inner.',
-    )
```
theo-brown, line 56: Was there a reason this was here rather than in the Pydantic config?

torax/_src/transport_model/transport_model.py
```diff
@@ -124,57 +131,59 @@ def __eq__(self, other) -> bool:
...
-    d_face_el = jnp.where(
-        mask,
-        dynamic_runtime_params_slice.transport.D_e_min,
-        d_face_el,
-    )
-    v_face_el = jnp.where(
-        mask,
-        0.0,
-        v_face_el,
-    )
```
theo-brown, line 176: @jcitrin is there a physical/numerical reason for setting `v_face_el` in the pedestal to 0.0 rather than `ve_min`?

torax/_src/transport_model/runtime_params.py
theo-brown: Oh this is probably the cause of the failing tests, nvm

torax/_src/transport_model/combined.py
```diff
@@ -0,0 +1,111 @@
...
+from torax._src.config import runtime_params_slice
+from torax._src.geometry import geometry
+from torax._src.pedestal_model import pedestal_model as pedestal_model_lib
+from torax._src.transport_model import runtime_params as runtime_params_lib
+from torax._src.transport_model import transport_model
+
+
+@chex.dataclass(frozen=True)
+class DynamicRuntimeParams(runtime_params_lib.DynamicRuntimeParams):
+  transport_model_params: Sequence[runtime_params_lib.DynamicRuntimeParams]
```
theo-brown, line 38: Do we need more strict typing here?

torax/_src/transport_model/tests/combined_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+  def test_call_implementation(self):
+    config = default_configs.get_default_config_dict()
+    config['transport'] = {
+        'model_name': 'combined',
+        'transport_models': [
+            {'model_name': 'constant', 'rho_max': 0.5, 'chi_i': 2.0},
+            {'model_name': 'constant', 'rho_min': 0.5, 'chi_i': 1.0},
+        ],
+    }
```
theo-brown, line 79: Would it be better if this were passed by a `@parameterized` decorator?

docs/configuration.rst
```diff
@@ -676,50 +688,77 @@ transport model.
 ``V_e_max`` (float [default = 50.0])
   Upper allowed bound for particle convection :math:`V`, in units of :math:`m^2/s`.
 
+``smoothing_width`` (float [default = 0.0])
+  Width of HWHM Gaussian smoothing kernel operating on transport model outputs.
+  If using the ``QLKNN_7_11`` transport model, the default is set to 0.1.
+
+``smooth_everywhere`` (bool [default = False])
+  Smooth across entire radial domain regardless of inner and outer patches.
```
theo-brown, line 696: Moved this so deprecated options come after

torax/tests/test_data/test_combined_transport.py
theo-brown, line 1: Potential to adjust the parameters here so that the results mean something physical. Currently, compared to `test_bohmgyrobohm_all` (R), this `combined` integration test looks like this (L):
image

As expected, there's elevated transport in the core from the inner constant term, spikes in intermediate values of rho from the CGM, and general BgB shape. Outer constant term gets entirely squashed by the pedestal, which by default sets transport outside of 0.91 to the minimum (post smoothing).

docs/configuration.rst
```diff
@@ -676,50 +688,77 @@ transport model.
 ``V_e_max`` (float [default = 50.0])
   Upper allowed bound for particle convection :math:`V`, in units of :math:`m^2/s`.
 
+``smoothing_width`` (float [default = 0.0])
+  Width of HWHM Gaussian smoothing kernel operating on transport model outputs.
+  If using the ``QLKNN_7_11`` transport model, the default is set to 0.1.
```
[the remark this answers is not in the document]
theo-brown: This was copied wholesale from earlier. It looks like the default is always 0.0, https://github.com/google-deepmind/torax/blob/2b82c3d631d0645b82b27e0d0b5a7138ffccadd0/torax/_src/transport_model/pydantic_model_base.py#L101

but I don't know whether QLKNN711 does anything to alter this.

torax/_src/transport_model/combined.py
theo-brown: I didn't know whether `Sequence[Union[QLKNNDynamicRuntimeParams, BohmgyrobohmDynamicRuntimeParams, ...]]` was 'more proper' or whether it's ok to say it can be a sequence of any transport model dynamic runtime params.

I think for the moment letting it be any is ok, because otherwise we'd be imagining a use case that doesn't exist yet (e.g. what if there's a transport model that isn't compatible with Combined?)

torax/_src/transport_model/combined.py
```diff
@@ -0,0 +1,111 @@
...
+          geo,
+          transport_component,
+          pedestal_model_outputs,
+      )
+
+      summed_transport_coeffs = jax.tree.map(
+          lambda old, new: old + new,
+          summed_transport_coeffs,
+          transport_component,
+      )
```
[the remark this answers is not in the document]
theo-brown: Oh cool I didn't realise that you do this with just a vanilla Python sum, I thought you had to jnp.stack and jnp.sum.

torax/_src/transport_model/runtime_params.py
```diff
@@ -54,3 +56,8 @@ def __post_init__(self):
         self.rho_outer <= self.rho_inner,
         'rho_outer must be greater than rho_inner.',
     )
+    jax_utils.error_if(
+        self.rho_max,
+        self.rho_max <= self.rho_min,
+        'rho_max must be greater than rho_min.',
+    )
```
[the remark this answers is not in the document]
theo-brown: Added one to catch both of these checks

theo-brown:
@tamaranorman thanks for the comments. Main response in https://github.com/google-deepmind/torax/pull/1233/commits/bd099a219488661118d75a8903bf40d0537a282d - I think I agree with all of your suggestions except the one about default smoothing width. If you're happy with my edits then I'll squash b099a2 and 9145bec together.

Changes to `_call_implementation` in a separate commit for clarity. https://github.com/google-deepmind/torax/pull/1233/commits/6c1d150f7984054844d7bd8c2a490d2c7eceb712.

jcitrin (commented):
Great PR! Thanks for all these efforts

theo-brown:
Updated integration test:
- Added smoothing on the combined coefficients, as it's a good opportunity to demo it
- Overwrote defaults for constant transport model
image

docs/scripts/combined_transport_example.py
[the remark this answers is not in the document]
tamaranorman: That makes sense, can you use the examples/basic_config instead of the test_utils? Just so we avoid using test code outside of tests

torax/_src/transport_model/bohm_gyrobohm.py
```diff
@@ -54,10 +54,11 @@ def __init__(
 
   def _call_implementation(
       self,
+      transport_dynamic_runtime_params: DynamicRuntimeParams,
```
tamaranorman, line 57: I have not checked but am not convinced this will type check as it is being based a list of the base classes. Either check the type in combined/base class or at the top of call_implementation

docs/configuration.rst
```diff
@@ -696,8 +720,12 @@ transport model.
   :math:`\hat{\rho}` below which inner patch is applied.
 
 ``apply_outer_patch`` (**time-varying-scalar** [default = False])
-  If ``True``, set a patch for outer core transport coefficients above ``rho_outer``.
-  Useful for the L-mode near-edge region where models like QLKNN10D are not applicable. Only used if ``set_pedestal==False``.
+  If ``True``, set a patch for outer core transport coefficients above
+  ``rho_outer``. Useful for the L-mode near-edge region where models like
+  QLKNN10D are not applicable. Only used if ``set_pedestal==False``.
+  If using a `CombinedTransportModel`, ensure that the inner patch is
```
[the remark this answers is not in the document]
theo-brown: Functionally there's no reason not to, but I agree it's not in the spirit of where the changes are heading. 
In the meantime, however, I think it's useful to still be here as it means you can do minor adjustments to existing configs to make them combined rather than having to convert patches to combined constants. Once we announce that patches are deprecated then we can update this.

jcitrin (approved):

theo-brown:
@jcitrin just fixing the conflicts with main - should I squash into a single commit or leave the history?

jcitrin:
Thanks. Can squash the commits

theo-brown:
Cool, ready to go from my end. Thanks for your feedback!

torax/_src/transport_model/bohm_gyrobohm.py
tamaranorman: Yes that is what I was meaning, can you change the type of this as well to the base class not the inherited to avoid errors

theo-brown:
* Fixed a duplicate line in the docs
* Added the integration test to the registry (forgot to do this before!)

PR #1296, opened 2025-06-18
group simulation preparation functionality in single method

mikesndrs:


jcitrin:
Merged with ac489ec , which included some minor modifications from the original PR:

* Linting cleanups
* More typehints
* Docstring extension

Great to have a first ITER/IC PR in through the door 🚀

Issue #941, opened 2025-04-03
Add pedestal region transport coefficient values to the pedestal config, instead of using the minimum values


jcitrin:

jcitrin:
Yes that sounds great.

theo-brown:
We discussed the following:

```
transport: {
     'model_name': 'combined',
     'core_transport_models': [...],
     'pedestal_transport_models': [...],
     'smoothing_param': ...,
     'clipping_param': ...,
  }
```

Is this still ok or would you rather the functionality be available to existing transport models without having to switch to `combined`? 

Looking ahead to our proposed  design for v2, I think my suggestion above most closely replicates it in a v1-compatible way, as v2 would then basically entail just removing `'model_name': 'combined'` (ie removing one level of nesting in the Pydantic config). Let me know what you think :)

jcitrin:
Discussed also with @tamaranorman. 

We agree that it's OK for the pedestal transport functionality to be limited to 'combined' for v1.

Since without combined, in v1 we'll still have 'transport_model' as a required key dict, it's then a bit jarring/confusing if we have both 'transport_model' and 'pedestal_transport_models'

jcitrin:
@aaronkho , see this PR as well as #1256 and #1233 . These all significantly help with the implementation of your exponential pedestal model.

aaronkho:
Since these other PRs are now merged, I maybe should chime in here.

I had previously implemented my model directly inside the `TransportModel._apply_clipping()` function. With these additions, I thought it would make sense to switch over to putting the functional form in a new file, `_src/transport_model/exponential_pedestal.py`. Should this be a completely new model class or should it inherit from `ConstantModel`? Or should it not be a new model class at all, instead setting the coefficients of a `ConstantModel` with values from the functional form directly at the run config level.

However, since `_apply_clipping()` is still called, wouldn't this just replace any coefficients I place into that region beyond the internal BC?

theo-brown:
File location and name seems sensible. At a later date if/when we've implemented more options for pedestal transport we can think whether they deserve to be in a subpackage.

My opinion on what would be best is an entirely new class inheriting from TransportModel with custom `_call_implementation`. Maybe raise an error if apply clipping is set. 

Current plan for this issue is for the CombinedModel to have an altered `__call__` method so that the pedestal from the new pedestal_transport_model section of config does not get overwritten. This would then become the standard for v2. I think this would work with the above in a sensible way

aaronkho:
[...]

PR #1330, opened 2025-06-25
Add pedestal transport model section to CombinedTransportModel

theo-brown:
- [x] Add separate pedestal model to CombinedTransportModel
- [x] Update tests
- [x] Update docs 
- [x] Add test to check behaviour when no pedestal model supplied - intended behaviour is revert to chi_min, etc 

Implements #941


torax/_src/transport_model/combined.py
```diff
@@ -33,19 +35,72 @@
...
+    # Apply min/max clipping
+    transport_coeffs = self._apply_clipping(
+        transport_runtime_params,
+        transport_coeffs,
+    )
+
+    # In contrast to the base TransportModel, we do not apply patches
+
+    # Return smoothed coefficients if smoothing is enabled
+    # TODO: what should be done about masking? In base TransportModel, the
```
theo-brown: What should be done about masking? Need to compare this method with what was there before.

theo-brown: Resolved, see updated comment

torax/_src/transport_model/pydantic_model.py
```diff
@@ -368,50 +368,69 @@ class CombinedTransportModel(pydantic_model_base.TransportBase):
 
   Attributes:
     model_name: The transport model to use. Hardcoded to 'combined'.
-    transport_models: A sequence of transport models, whose outputs will be
-      summed to give the combined transport coefficients.
+    core_transport_models: A sequence of transport models, whose outputs will be
+      summed to give the combined core transport coefficients.
```
tamaranorman, line 372: This change is technically breaking for anyone with a config that uses a CombinedTransportModel and technically we have agreed not to change that. This was only recently added but we should decide if we think we want to allow this.

torax/_src/transport_model/pydantic_model.py
```diff
@@ -368,50 +368,69 @@ class CombinedTransportModel(pydantic_model_base.TransportBase):
...
+    pedestal_transport_model: A model for pedestal transport coefficients.
   """
 
-  transport_models: Sequence[CombinedCompatibleTransportModel] = pydantic.Field(
+  core_transport_models: Sequence[
+      CombinedCompatibleTransportModel
+  ] = pydantic.Field(
       default_factory=list
   )  # pytype: disable=invalid-annotation
+  pedestal_transport_model: CombinedCompatibleTransportModel
```
tamaranorman: Is this required? Or should we still support the old post pedestal model if none is provided using the coefficients? We have a default for the core_transport_models which seems to be an empty list? What is the point of that, or should they both be required

torax/_src/transport_model/pydantic_model.py
```diff
@@ -368,50 +368,69 @@ class CombinedTransportModel(pydantic_model_base.TransportBase):
...
+
     return combined.DynamicRuntimeParams(
-        transport_model_params=model_params_list,
+        core_transport_model_params=core_transport_model_params,
+        pedestal_transport_model_params=pedestal_transport_model_params,
         **base_kwargs,
     )
 
   @pydantic.model_validator(mode='after')
   def _check_fields(self) -> typing_extensions.Self:
```
tamaranorman, line 421: Do we need some validation on the pedestal model inputs

docs/scripts/combined_transport_example.py
```diff
@@ -51,13 +51,13 @@ def main(argv: Sequence[str]) -> None:
...
-                  'chi_i': 0.5,
-                  'rho_min': 0.5,
-                  'rho_max': 1.0,
-              },
           ],
+          'pedestal_transport_model': {
+              'model_name': 'constant',
+              'chi_i': 0.5,
+              'rho_min': 0.5,
+              'rho_max': 1.0,
```
tamaranorman: Should rho be set or should we just have it be everything after the pedestal?

docs/scripts/combined_transport_example.py
```diff
@@ -51,13 +51,13 @@ def main(argv: Sequence[str]) -> None:
                   'rho_min': 0.2,
                   'rho_max': 0.5,
```
tamaranorman: Do we need errors for regions not covered by a transport model?

docs/scripts/combined_transport_example.py
tamaranorman, line 1: Was this not used to generate something in the docs, does that now need updating

torax/_src/transport_model/combined.py
```diff
@@ -100,20 +155,63 @@ def _call_implementation(
...
 
   def __eq__(self, other):
     return (
         isinstance(other, CombinedTransportModel)
-        and self.transport_models == other.transport_models
+        and self.core_transport_models == other.core_transport_models
+        and self.pedestal_transport_model == other.pedestal_transport_model
+    )
+
+  def _apply_pedestal_domain_restriction(
```
tamaranorman, line 181: nit: Move logic above the hash and equals models to keep them altogether

docs/scripts/combined_transport_example.py
jcitrin: It's a valid option in the sense of it's "OK" for a zone to have no turbulent transport defined. That zone will still have set the minimum transport coefficient values, and in the PDE itself (in calc_coeffs.py) there may be the addition of neoclassical transport (if on). While there's only limited use-cases of having no turbulent transport in a specific zone (maybe internal-transport-barrier explorations?), I don't think we need errors if this occurs

jcitrin: That being said I'm not sure that this is a good example. rho_max should be 1.0, and the pedestal model will supersede it in the pedestal region. See comment below

docs/scripts/combined_transport_example.py
jcitrin: Agree. A pedestal transport model should be in the pedestal zone as set by the 'pedestal' config. We shouldn't have two separate definitions of where the pedestal lives. pedestal_transport_models thus should not have a rho_min and rho_max.

torax/_src/transport_model/combined.py
```diff
@@ -100,20 +155,63 @@ def _call_implementation(
...
+        pedestal_transport_coeffs,
+        pedestal_model_output,
+    )
 
+    # Combine the transport coefficients from core and pedestal models.
     combined_transport_coeffs = jax.tree.map(
         lambda *leaves: sum(leaves),
-        *component_transport_coeffs_list,
+        *(core_transport_coeffs_list + [pedestal_transport_coeffs]),
     )
```
jcitrin: There should be symmetry with core_transport_coeffs in the sense that multiple pedestal_transport_models should also be allowed

torax/_src/transport_model/combined.py
jcitrin, line 181: Doesn't need to be for this PR, but it feels wasteful to run a pedestal model for the entire radius, when in fact most of the indices will be set 0, since the pedestal region is small.

We may want to refactor transport models to only do calculations within the domain of interest. Likely this can also be made vmappable etc. Can add a TODO to "combined"

torax/_src/transport_model/pydantic_model.py
jcitrin: agree. Should revert this. It's ok to have "transport_models" and "pedestal_transport_models". In V2 we will remove combined altogether by essentially making it default, and then having "core_transport_models" and "pedestal_transport_models"

torax/_src/transport_model/pydantic_model.py
jcitrin: Agree. If None is provided for pedestal_transport_model, should revert to the De_min, chi_min default.

jcitrin (commented):
In addition to the comments, please update the docs following the converged-upon patterns.

torax/_src/transport_model/pydantic_model.py
theo-brown: Sure - I had forgotten that Combined had already made its way into a minor patch. Will revert and add a note about v2

torax/_src/transport_model/pydantic_model.py
```diff
@@ -368,50 +368,69 @@ class CombinedTransportModel(pydantic_model_base.TransportBase):
...
     super()._check_fields()
-    if not self.transport_models:
+    if not self.core_transport_models:
       raise ValueError(
-          'transport_models cannot be empty for CombinedTransportModel. '
+          'core_transport_models cannot be empty for CombinedTransportModel. '
           'Please provide at least one transport model configuration.'
       )
```
theo-brown: As per Jonathan's comment, there is a conceivable scenario where only neoclassical transport is used and no turbulent transport. Hence, removing this as an unnecessary check.

docs/scripts/combined_transport_example.py
theo-brown: Yep, but was just leaving docs changes until source code changes had finalised!

torax/_src/transport_model/combined.py
theo-brown: Agreed, this was a very botched way of doing it. My reasoning was trying to avoid creating a new base class for PedestalTransportModels and so it made most sense to me to just replicate what was done in TransportModel. I'll add a comment that this should be replaced with something more efficient.

theo-brown: On the topic of vmappability - is switching all TransportModels to be vmapped over their domain going to cause problems with different surrogates people may build? My concern is to do with management of state and stuff, like whether to split RNGs or not. It makes total sense for the deterministic physics calculation models (e.g. BgB), but could it be a potential footgun for surrogates?

Could be trivial to navigate but would have to be documented as a requirement for models _call_implementation.

theo-brown: Also, are you suggesting vmapping over the domain or over the models? Not sure we could vmap over heterogeneous structures (ie, different models)

jcitrin: Not sure about stochastic surrogates. Of course surrogates can also be deterministic (and the ones we use currently are). It's a subclass of surrogates that would include stochastic elements. Instinct is "fix it when it's a problem", there are likely solutions.

On the mapped index, most natural one for now corresponds to the radial coordinate.

torax/_src/transport_model/pydantic_model.py
theo-brown: Added a test to check for this behaviour.

torax/_src/transport_model/combined.py
theo-brown: I've done a small refactor to improve code reuse. I think your suggested change is good but it should be in another PR? Made an issue, #1347

torax/_src/transport_model/pydantic_model.py
theo-brown: Added checks that patches, rho_min, and rho_max are not set on pedestal models

tamaranorman (approved):

docs/configuration.rst
```diff
@@ -1127,8 +1141,8 @@ This would produce a ``chi_i`` profile that looks like the following.
 Note that in the region :math:`[0, 0.2]`, only the first component is active,
 so ``chi_i = 1.0``. In :math:`(0.2, 0.3]` the first two components are both
 active, leading to a combined value of ``chi_i = 3.0``. In :math:`(0.3, 0.5]`,
-only the second model is active (``chi_i = 2.0``), and in :math:`(0.5, 1.0]`
-only the fourth model is active (``chi_i = 0.5``).
+only the second model is active (``chi_i = 2.0``), and in :math:`(0.8, 1.0]`
+only the pedestal transport model is active (``chi_i = 0.5``).
```
jcitrin: This is incorrect. From rho 0.3 to the default rho_norm_ped_top of 0.91, only the second transport_model is active. Then, from the default rho_norm_ped_top=0.91 to the edge, the pedestal model is active with chi_i=0.5.

Recommend explicitly setting `rho_norm_ped_top` in 'pedestal' to some value like 0.9 to make the zones more explicitl

docs/scripts/combined_transport_example.py
```diff
@@ -33,7 +35,12 @@ def main(argv: Sequence[str]) -> None:
           'geometry_type': 'circular',
           'n_rho': 30,  # for higher resolution plotting
       },
-      'pedestal': {},
+      'pedestal': {
```
jcitrin, line 37: Recommend setting `rho_norm_ped_top` here

jcitrin (approved):
Looks good! Following the small fixes + fixing the merge conflicts, will bring it it

docs/configuration.rst
theo-brown: Good catch, not sure why I made this mistake! Added `rho_norm_ped_top` and changed the explanation to fix. I regenerated the plot but it still looks the same, presumably because the profile is on the discrete rho grid and thus can't resolve the difference between ρ=0.91 and ρ=0.9.

theo-brown:
I've made the changes, rebased onto main, and squashed into a single commit, let me know if there's anything more that needs doing!

jcitrin (approved):

PR #1371, opened 2025-07-17
Validate electron cycotron heating and current drive module

theo-brown:
Aim: validate the implementation of the Lin-Liu EC model against JETTO

- [x] Correct the current drive and efficiency formulae (reflecting ongoing work in checking the implementation in JETTO) 
- [ ] Replicate simple EC-based TORAX scenarios in JETTO
  - [ ] STEP
  - [ ] Other? Good to have a non-spherical example as well
- [ ] Update docs and derivation document


theo-brown:
I've updated the `iterhybrid_predictor_corrector_ec_linliu` test case to use the current drive efficiency from Simon's paper, ζ = 0.35, with the corrected current drive formula.

Difference in EC current drive probably due to the fact that the new implementation in TORAX doesn't make the assumption that <1/R> ≈ 1/R, in order to support ST plasmas. 

In this picture, (1) is new, (2) is old.
Figure_1

I'm not sure whether the goal with this test is to replicate Simon's paper exactly? If so, I think we've got the EC targeted in slightly the wrong place, as it should be at ρ=0.33. However, the new model does appear to match the q profile better, as it's less reversed.

torax/_src/sources/electron_cyclotron_source.py
```diff
@@ -92,32 +94,51 @@ def calc_heating_and_current(
...
+  j_tor_ec = (
+      16
+      * jnp.pi
+      * constants.CONSTANTS.epsilon0**2
+      * core_profiles.T_e.value
+      * 1e3  # T_e in eV
+      * dynamic_source_runtime_params.current_drive_efficiency
+      * ec_power_density
+      / (
+          constants.CONSTANTS.qe**2
```
theo-brown: Would it be good to do this via the log for stability?

torax/_src/sources/electron_cyclotron_source.py
```diff
@@ -92,32 +94,51 @@ def calc_heating_and_current(
...
+      * fsa_1_R
+      * (
+          1
+          + jnp.where(
+              q_cell == 0,
+              jnp.zeros_like(geo.rho_norm),
+              geo.g2
+              * geo.g3
+              / (16 * jnp.pi**4 * q_cell**2 + constants.CONSTANTS.eps),
+          )
```
theo-brown: This was sort of a hack. Is there a better way around it?

torax/_src/sources/electron_cyclotron_source.py
```diff
@@ -140,18 +161,21 @@ class ElectronCyclotronSourceConfig(base.SourceModelBase):
...
+      grid.
+    gaussian_width: Gaussian EC power density profile width.
+    gaussian_location: Gaussian EC power density profile location.
+    P_total: Gaussian EC total power.
   """
 
   model_name: Literal["gaussian_lin_liu"] = "gaussian_lin_liu"
   current_drive_efficiency: torax_pydantic.TimeVaryingArray = (
-      torax_pydantic.ValidatedDefault({0.0: {0.0: 0.2, 1.0: 0.2}})
+      torax_pydantic.ValidatedDefault({0.0: {0.0: 0.25, 1.0: 0.25}})
```
theo-brown, line 162: Adjusted to better reflect common operating conditions under new scaling

torax/_src/sources/electron_cyclotron_source.py
jcitrin: yes wouldn't hurt, for when going to fp32 (GPU)

jcitrin:
> I'm not sure whether the goal with this test is to replicate Simon's paper exactly? 

No, this wasn't the point. We were not trying to replicate the paper exactly.

torax/_src/sources/electron_cyclotron_source.py
jcitrin: Which part is the hack?

Edit: I guess you mean the jnp.where with the q_cell==0 branch? If so, then actually I'm a bit confused as to when that branch will be hit. When are the source functions ever called with a non-defined pre-initialized core_profiles?

torax/_src/sources/electron_cyclotron_source.py
```diff
@@ -92,32 +94,51 @@ def calc_heating_and_current(
...
+          )
+          * core_profiles.n_e.value  # n_e in m^-3
+      )
+  )
+
+  # < 1/R > on cell grid, needed to convert j_tor to < j.B >
+  # TODO: add to geometry.Geometry?
+  # Numerical trick to avoid division by zero: dρ/dA = dρ²/dA * 1/(2ρ)
+  darea_drho2 = jnp.diff(geo.area_face) / jnp.diff(geo.rho_face**2)
+  fsa_1_R = geo.F * geo.g3 * darea_drho2 / (jnp.pi * geo.B_0)
```
jcitrin: where is this nice formula from?

theo-brown: Given

image

we have $\frac{\partial{A}}{\partial\rho} = \frac{1}{2\pi} \frac{\partial V}{\partial \rho} \left\langle\frac{1}{R}\right\rangle$, by evaluating the above with $f=1$ and then taking the partial derivative wrt $\rho$ (and recalling that the notation here uses the physicist's way of writing integrals, `∫dV f = ∫fdV`). This is taken from WIP JETTO docs, using $S$ to represent poloidal cross sectional area.

I reimplemented $\frac{\partial V}{\partial \rho}$ because there's some numerical issue with how it's computed in `StandardGeometry` that I haven't identified yet that crops up when you run STEP geometries:

For the ITER-hybrid case, `geo.vpr` is fine,
image
but in STEP SPR45, there's something wrong:
image

In these plots, I used `vpr_new =  4 * jnp.pi * geo.Phi_b * geo.rho_norm / (geo.F * geo.g3)`.

However, I've just noticed that my implementation in the PR is actually incorrect because I've not been careful about mixing my `rho` and `rho_norm`s! Thanks for bringing this to my attention, I'll fix it soon.

theo-brown: Cool, revisited with fresh eyes and I think it's better to just do `fsa_1_over_R = 2 * jnp.pi * jnp.gradient(geo.area, geo.volume)`. It's both simpler and I think more accurate - see below, where `<1/R>` should be a constant.
image

theo-brown: Ok, not sure what the best fix is here. Might need to wait until #1391 is sorted to be able to check <1/R> is correct.

jcitrin: Ah yes, thanks for reminding me of the formula. It's also in [Federico's thesis](https://infoscience.epfl.ch/server/api/core/bitstreams/10b008ee-a9c5-49ba-9fa5-a52fc3e69669/content) equation 6.19

jcitrin: > but in STEP SPR45, there's something wrong:

Note the near-axis smoothing for rho<0.1 which may be playing a role here, particularly when doing the tracing of an EQDSK file which may have some tracing issues in the more sparse axis region. I don't think missing a bit near-axis should impact a scenario much. https://github.com/google-deepmind/torax/blob/main/torax/_src/geometry/standard_geometry.py#L246

theo-brown: Yes, this may well be the cause. In this particular example it made quite a big difference to the EC efficiency and therefore q profile for rho<0.1 which is one of the main things we care about!

torax/_src/sources/electron_cyclotron_source.py
theo-brown: Added in latest commit

torax/_src/sources/electron_cyclotron_source.py
theo-brown: I've been trying to chase this down. I currently have no idea why this works.

To make things easier to debug, I've switched my config to use the linear solver and set a much smaller timestep (1e-3 compared to 5 with the Newton-Raphson). With the `jnp.where(q==0)` guard it solves fine, without it the solver produces nans in all the core profiles for the first timestep.

If I disable predictor-corrector (while keeping linear and small dt) it solves ok without the guard.

jcitrin: We can trace through the calculation at your next visit and see why this is needed and if a refactor would help avoid it

theo-brown: Thanks for the help tracing it today! #1460 for reference.

theo-brown:
Couple of updates following work today:
- Should I add a derivation of the equations somewhere? We've got an internal document at UKAEA that outlines everything, but there's a lot of other stuff in it about JETTO internals that is not relevant. Could maybe summarise the key points and add to docs, replacing the existing EC derivation?
- Happy to wait for #1460 to be merged, then can remove the machine epsilon guard
- This is also blocked by #1459, which should be fairly easy to fix

jcitrin:
Sorry for delay in replying.

> Should I add a derivation of the equations somewhere? We've got an internal document at UKAEA that outlines everything, but there's a lot of other stuff in it about JETTO internals that is not relevant. Could maybe summarise the key points and add to docs, replacing the existing EC derivation?

Yes this sounds great

> Happy to wait for https://github.com/google-deepmind/torax/pull/1460 to be merged, then can remove the machine epsilon guard

Now merged. Can remove the epsilon guard.

> This is also blocked by https://github.com/google-deepmind/torax/issues/1459, which should be fairly easy to fix

This can be fixed in parallel and is not a blocker per se. When doing the validation, can always modify the JETTO output by B0/<Btotal> , and then revalidate again later when the true B is used in TORAX.

theo-brown:
Rebased onto main and removed the epsilon guard.

I'm going to split off validation into a separate PR, which should enable this to be merged and then the validation done separately as part of the wider ST validation. This PR is just implementing the corrected ST-compatible model.

torax/_src/physics/collisions.py
```diff
@@ -182,6 +183,28 @@ def fast_ion_fractional_heating_formula(
...
+    n_e: Electron density in m^-3.
+
+  Returns:
+    Coulomb logarithm.
+  """
+  # Rescale T_e to eV for specific form of formula.
+  T_e_ev = T_e * 1e3
+  return 31.0 - 0.5 * jnp.log(n_e) + jnp.log(T_e_ev)
+
+
```
jcitrin, line 207: I thought the difference between the ee and ei formulas (see below this one) was a bit larger (just a bit) than 1% in the leading constant. Won't change much but worth double checking.

torax/tests/test_data/test_iterhybrid_predictor_corrector_ec_linliu.nc
jcitrin, line 1: how different are the results?

jcitrin (commented):
I'm happy to bring this in, but is there any validation plot to JINTRAC that could be added to the PR just to demonstrate the correctness of the equations?

Also, need to sync to main and also squash commits (I forgot to remind to squash before the other ones that have already started the import process, no worries about those)

theo-brown:
### Comparison with JINTRAC

I took profiles from an ST JINTRAC run, including QECE and JZEC profiles.
JZEC converts to <j.B> via
image
where A = <R^-2> and K=<|grad V|^2 / R^2> (as in JETTO outputs), although apparently JETTO approximates this by dropping the second term in the brackets (?).

JETTO does a collection of weird things to represent the EC efficiency, so I just computed the efficiency from the JZEC and QECE profiles using
image
using the conversion 
image


In the following, solid is a TORAX run with the ECRH module, current evolution only. Dashed is a TORAX run with prescribed external current source set to <j.B>/B0 from JZEC, current evolution only.

Figure_1




image

torax/_src/physics/collisions.py
theo-brown: From Wesson:
image

theo-brown: Is there a better source that gives a larger discrepancy?

torax/tests/test_data/test_iterhybrid_predictor_corrector_ec_linliu.nc
theo-brown: Figure_1

theo-brown:
Squashed and rebased!

theo-brown:
Sorry, please ignore my re-request for review - still working on the JINTRAC comparison, found a problem with my efficiency calculation.

theo-brown:
@jcitrin as requested!

### Changes

- Squashed commits
- Rebased onto main

### Comparison with JETTO + GRAY for EC

- Used a JETTO GRAY run to compute $\zeta := \frac{e^3 \ln\Lambda}{16 \pi \varepsilon^2 } \frac{n_e \mathrm{[m^{-3}]}}{T_e \mathrm{[J]}}\ \frac{j_\mathrm{tor} \mathrm{[Am^{-2}]}}{Q_{ec} \mathrm{[Wm^{-3}]}}$
- Input ζ profile as `current_drive_efficiency` and the EC power density profile from the JETTO+GRAY run as `extra_prescribed_power_density` for the `ecrh` source
- Ran with Vloop boundary condition, evolving current equation only, cranked up resistivity multiplier to be able to simulate for shorter times
- Plotted profiles from the JETTO JSP output and the TORAX NetCDF output. EC current profiles match exactly.

ec_validation

torax/tests/test_data/test_iterhybrid_predictor_corrector_ec_linliu.nc
theo-brown: Oh sorry did you want a % difference?

jcitrin: This is fine, thanks!

jcitrin (approved):

jcitrin:
Sorry that it took so long to get back on this. Can you please merge with head again? I'll then bring it in

PR #1413, opened 2025-08-04
Add <1/R> to geometry and calulate spr from <1/R>

theo-brown:
- [x] Add to geometry
- [x] Add to CHEASE loader
- [x] Add to IMAS loader
- [x] Add to LY loader
- [x] Add to EQDSK loader
- [x] Handle division by zero
- [x] Rerun tests and sense check the changes to outputs


jcitrin:
@felicif , @theo-brown is carrying out the an extension of the geometry inputs to include the 1/<R> term. Where can this be found in the LY structure?

EDIT: nvm, it's Q0Q

theo-brown:
### Effect of changes on test cases

TLDR: Minor differences in j profiles (and hence Ibs), De, Ve, Qalpha (and hence P alpha). Differences can be pronounced in chi, but only for one or two timesteps before returning to better agreement. These differences tend to not make a difference to overall behaviour of the simulation. Sawteeth behaviour has been altered, but I need a second pair of eyes to ascertain whether it's ok.

#### iterhybrid_predictor_corrector cases
Observe differences in chi, D_e, j, Qalpha, but without any real impact on the wider behaviour.
test_iterhybrid_predictor_corrector

#### iterhybrid_rampup cases
Observe how De is very different for one timestep but then returns to agreement for the next timestep. 
test_iterhybrid_rampup_78
test_iterhybrid_rampup_80

#### test_iterhybrid_rampup_sawtooth
More pronounced flattening effect in the core. @jcitrin is this within the margin of acceptable changes to tests?
test_iterhybrid_rampup_sawtooth

theo-brown:
- A few tests are failing because they compare against `torax._src.test_utils.torax_refs`. What's the correct way of updating this file?
- I also don't quite understand the following check that is now failing:
https://github.com/google-deepmind/torax/blob/ec1988b676a8dc3a98c362f61ed807da9693b40c/torax/_src/core_profiles/tests/initialization_test.py#L321-L325
Should this hold because the total plasma current is an externally imposed constraint?
- Finally, as there's no LY file bundled with TORAX I can't fully check that the loading is ok. Is Q0Q always present, or should I add a check like in the IMAS case where gm9 is optional?

jcitrin:
> More pronounced flattening effect in the core. @jcitrin is this within the margin of acceptable changes to tests?

Regarding the sawteeth, it's probably just because the sawtooth crash times are now slightly different, and in that specific time slice you see one simulation post-crash and the other is more recovered. Here what matters more is the average over time rather than the difference at one time slice. e.g. if you plot Te[0] of simulations over time, as well as the average difference over multiple sawtooth crashes. If the compare_sim_test scripts are showing small differences for the averaging, then it should be fine.

jcitrin:
> Observe how De is very different for one timestep but then returns to agreement for the next timestep.

Likely related to stiffness and very slight profile differences triggering for one time slice a seemingly large difference. Over time the averages should be quite similar

jcitrin:
> A few tests are failing because they compare against torax._src.test_utils.torax_refs. What's the correct way of updating this file?

We have an internal notebook to regenerate those, which I think should be exposed if you need it. However, the quickest would be to just leave it for now and we will regenerate those refs during internal review. In due course we'll make the regeneration script public and provide instructions on its use

jcitrin:
> I also don't quite understand the following check that is now failing:
> Should this hold because the total plasma current is an externally imposed constraint?

Strange that it's failing. By how much? However look again I think that comparing the mean is a bit weird. We should be comparing all elements the cumulative integral (using spr as the metric coefficient), i.e. the plasma current up to a given flux surface. That should hold since the only components of plasma current in the test case is Ohmic and bootstrap. Feel free to change it, otherwise can leave it and we'll look at it during internal review

jcitrin:
> Finally, as there's no LY file bundled with TORAX I can't fully check that the loading is ok. Is Q0Q always present, or should I add a check like in the IMAS case where gm9 is optional?

Q0Q should always be present

theo-brown:
>  > Observe how De is very different for one timestep but then returns to agreement for the next timestep.
>
> Likely related to stiffness and very slight profile differences triggering for one time slice a seemingly large difference. Over time the averages should be quite similar

Verified with taking a moving average in time over the D_e profiles, the averaged profiles end up matching very well except for two timesteps (a difference which decreases as you increase the width of the averaging window).

```python
from scipy.signal import savgol_filter
import matplotlib.cm as cm

colors = cm.viridis(np.linspace(0, 1, len(original.time)))

# Average over time
smoothed_modified_D_turb_e = savgol_filter(
    modified.profiles.D_turb_e, window_length=5, polyorder=0, axis=0
)
smoothed_original_D_turb_e = savgol_filter(
    original.profiles.D_turb_e, window_length=5, polyorder=0, axis=0
)
for t in range(41):
  plt.plot(smoothed_modified_D_turb_e[t], color=colors[t])
  plt.plot(smoothed_original_D_turb_e[t], "--", color=colors[t])

plt.plot([], [], "k", label="Modified")
plt.plot([], [], "k--", label="Original")
plt.legend()
```
image

theo-brown:
> Regarding the sawteeth, it's probably just because the sawtooth crash times are now slightly different

Absolutely correct, verified using same method as above and more careful manual inspection of the nonsmoothed timetraces

image

theo-brown:
>>    I also don't quite understand the following check that is now failing:
>>    Should this hold because the total plasma current is an externally imposed constraint?
>
> Strange that it's failing. By how much? However look again I think that comparing the mean is a bit weird. We should be comparing all elements the cumulative integral (using spr as the metric coefficient), i.e. the plasma current up to a given flux surface. That should hold since the only components of plasma current in the test case is Ohmic and bootstrap. Feel free to change it, otherwise can leave it and we'll look at it during internal review

1. Ah, it's failing with a 0.0305 relative difference, and the tolerance is set to 0.03. That's not really worth losing sleep over imo, I think we can just bump the tolerance to 0.04?

2. There's a comment justifying the choice of comparing the mean in `test_compare_initial_currents_with_different_initial_j_ohmic` that says

https://github.com/google-deepmind/torax/blob/4b3c09c08ca0c0bb254da2fe166658a302a3f040/torax/_src/core_profiles/tests/initialization_test.py#L170-L172

To keep this PR well-scoped I might just boost the tolerance, and leave a note.

theo-brown:
>>  A few tests are failing because they compare against torax._src.test_utils.torax_refs. What's the correct way of updating this file?
>
> We have an internal notebook to regenerate those, which I think should be exposed if you need it. However, the quickest would be to just leave it for now and we will regenerate those refs during internal review. In due course we'll make the regeneration script public and provide instructions on its use

Remaining failing CI is only the tests from this file.

torax/_src/geometry/geometry.py
```diff
@@ -333,6 +337,24 @@ def g1_over_vpr2_face(self) -> jax.Array:
         [jnp.expand_dims(first_element, axis=-1), bulk], axis=-1
     )
 
+  @property
+  def gm9(self) -> jax.Array:
+    r"""<1/R> on cell grid [:math:`\mathrm{m}^{-1}`]."""
+    bulk = 2 * jnp.pi * self.spr[..., 1:] / self.vpr[..., 1:]
+    first_element = 1 / self.R_major
+    return jnp.concatenate(
+        [jnp.expand_dims(first_element, axis=-1), bulk], axis=-1
```
jcitrin: No need to concatenate for the cell grid, since the first element does not correspond to rho=0 . Also the first_element is thus incorrect. Simply do 

```
return 2 * jnp.pi * self.spr / self.vpr
```

jcitrin: unfortunately this may lead to tiny test errors again and a need to regenerate the nc files

jcitrin:
Looks great! Happy to bring it in after the one requested change.

theo-brown:
Thanks for the review @jcitrin. Made the change and updated the necessary test files (changes were order 1e-4). Reminder that the remaining tests failing are due to `torax/_src/physics/tests/psi_calculations_test.py` and the reference values in ` torax._src.test_utils.torax_refs`.

Issue #1391, opened 2025-07-28
Validate geometry loading for spherical tokamaks


theo-brown:
I think there's at least one assumption (maybe more) in the EQDSK loader (and maybe elsewhere) that breaks down for STs.

Examples:

$\frac{\partial V}{\partial \rho} = \frac{4\pi\Phi_b \hat{\rho}}{F \langle1/R^2\rangle}$:

<img width="571" height="435" alt="Image" src="https://github.com/user-attachments/assets/ac064312-e040-4e25-a807-69539ae44e22" />

$<1/R> = 2\pi \frac{\partial A}{\partial V}$ (JETTO vs TORAX, same eqdsk):

<img width="556" height="418" alt="Image" src="https://github.com/user-attachments/assets/95419c22-8944-43e0-8663-e2ba1435cbd2" />

Note <1/R> is not expected to be constant for STs.

jcitrin:
On the first example, as discussed in #1371, I think that may be an issue with the near-axis smoothing, exasperated by using an EQDSK file.

jcitrin:
On the second example, the issue is probably that we need to decouple the area and volume calculations. dA/dV = 1/(2piR) by construct in TORAX. e.g. [in from_eqdsk](https://github.com/google-deepmind/torax/blob/c9ab82ed4a2a2b778cc24637a5829bc71b86dc1f/torax/_src/geometry/standard_geometry.py#L822) , and in [build_standard_geometry](https://github.com/google-deepmind/torax/blob/c9ab82ed4a2a2b778cc24637a5829bc71b86dc1f/torax/_src/geometry/standard_geometry.py#L1046)

theo-brown:
Yep, agree on the second. For the first, I could test loading the same geometry from some other sources? I agree it looks like an interpolation/smoothing issue rather than something more serious. I think I can dig out a CHEASE STEP geometry and can easily get an IMAS one

jcitrin:
> For the first, I could test loading the same geometry from some other sources? I agree it looks like an interpolation/smoothing issue rather than something more serious. I think I can dog out a CHEASE STEP geometry and can easily get an IMAS one

That sounds good

theo-brown:
Regarding the point you mentioned about it being surprising how much the approximation `<1/R> ≈ 1/R` breaks down when we spoke in person.

I think the Shafranov shift is just big in STs like STEP, increasing by ~10x over the radius - see a comparison below between something STEPish and something JETish.

[...]

PR #1416, opened 2025-08-05
Add `slice_time` and `slice_index` arguments to IMAS geometry loader

theo-brown:
This enables loading an equilibrium from any timeslice in an equilibrium IDS. Useful if resuming from pre-existing simulations.


theo-brown:
@MateoBell what's the provenance of the `ITERhybrid_COCOS17_IDS_ddv4.nc` file? I'd like to extend it to have multiple time slices so that I can test what I've implemented in this PR.

torax/_src/geometry/tests/imas_test.py
```diff
@@ -98,6 +99,60 @@ def test_IMAS_input_with_equilibrium_object(self):
     )
     config.build_geometry()
 
+  def test_IMAS_loading_specific_slice(self):
+    filename = 'ITERhybrid_COCOS17_IDS_ddv4.nc'
+    SLICE_TIME = 5
+    SLICE_INDEX = 1
```
theo-brown: Need to find an IMAS .nc file with multiple slices

theo-brown:
Linked to #1391. During investigating the different geometry loaders in ST, I am comparing eqdsk from various timesteps with CHEASE, IMAS, etc. Hence, I need to be able to load different slices from the IMAS equilibrium IDS.

MateoBell:
> @MateoBell what's the provenance of the `ITERhybrid_COCOS17_IDS_ddv4.nc` file? I'd like to extend it to have multiple time slices so that I can test what I've implemented in this PR.

The .nc file was generated with chease converting the [EQDSK_ITERhybrid_COCOS02.eqdsk](https://github.com/google-deepmind/torax/blob/main/torax/data/third_party/geo/EQDSK_ITERhybrid_COCOS02.eqdsk) (actually not completely sure which eqdsk of the examples, but it is the one which was used to compare to the CHEASE default geometry file). There is a script to run CHEASE from an eqdsk and save it to an IDS in a database, and then I converted it in netCDF. The script probably requires IMAS core so I can try to generate an IDS with multiple time slices with given eqdsk if needed.

theo-brown:
Thanks! I think for this PR we ideally would have an IDS with different profiles across the time steps we're comparing. This will also pave the way for smoothly interpolating between geometries to have evolving geometry over the run, but this may require rerunning CHEASE (or equivalent fixed-boundary solver) to produce something consistent.

We could, for example, run CHEASE twice using the pprime and FFprime profiles from the beginning and end of the TORAX iterhybrid rampup case as inputs, to give two different geometries for the start and end of the ramp.

 What do you think?

MateoBell:
> We could, for example, run CHEASE twice using the pprime and FFprime profiles from the beginning and end of the TORAX iterhybrid rampup case as inputs, to give two different geometries for the start and end of the ramp.

Yes fully agree ! I think it is a good idea so we can get different geometries at different times to compare.

theo-brown:
I've never run CHEASE before, if you're able would you be able to generate the two geometries?

Failing CI is just the placeholder test that's waiting for the new data.

MateoBell:
Ok I think I should be able to do it, will try to do it asap !

jcitrin:
Thanks both! Having a multiple time-slice IDS loader and a test for it would be ideal

MateoBell:
I added the multiple time slices IDS, but I just noticed the netCDF file is pretty heavy. I generated it with CHEASE for all time slices of the output simulation from iterhybrid_rampup config. If needed I can reduce the number of time slices to reduce the weight. 
Also I modified a bit the tests to make sure the IDS was generated properly and passing it, I fixed it rapidly so you might want to do it more properly.

theo-brown:
Incredible, thanks so much @MateoBell. Your fixes to the tests also seem fine!

In the latest force push I just rebased onto main.

For reference, the full .nc is 56.21 MB (GitHub's recommended limit is 50MB). I think we can definitely reduce the number of timeslices - maybe we should go down to ~10 from ~40?

jcitrin:
Thanks both!

> ~10 from ~40?

Yes, good idea to not load up the repo with large files

MateoBell:
Okay just replaced it with 11 time slices evenly spaced in time, the file is now ~16MB !

torax/_src/geometry/pydantic_model.py
```diff
@@ -300,6 +303,8 @@ class IMASConfig(torax_pydantic.BaseModelFrozen):
   imas_filepath: str | None = 'ITERhybrid_COCOS17_IDS_ddv4.nc'
   imas_uri: str | None = None
   equilibrium_object: ids_toplevel.IDSToplevel | None = None
+  slice_index: int = 0
```
jcitrin: make it a pydantic.PositiveInt

torax/_src/geometry/imas.py
```diff
@@ -71,9 +77,24 @@ def geometry_from_IMAS(
     raise ValueError(
         "equilibrium_object must be a string (file path) or an IDS"
     )
-  # TODO(b/431977390): Currently only the first time slice is used, extend to
+  # TODO(b/431977390): Currently only a single time slice is used, extend to
   # support multiple time slices.
-  IMAS_data = equilibrium.time_slice[0]
+  if slice_time is not None:
+    if slice_time in equilibrium.time:
+      slice_index = np.nonzero(equilibrium.time == slice_time)[0]
```
jcitrin: maybe this should check up to a tolerance to have some flexibility to rounding errors?

torax/_src/geometry/pydantic_model.py
```diff
@@ -300,6 +304,8 @@ class IMASConfig(torax_pydantic.BaseModelFrozen):
   imas_filepath: str | None = 'ITERhybrid_COCOS17_IDS_ddv4.nc'
   imas_uri: str | None = None
   equilibrium_object: ids_toplevel.IDSToplevel | None = None
+  slice_index: pydantic.NonNegativeInt = 0
+  slice_time: float | None = None
```
Nush395: should this be a positive float?

torax/_src/geometry/pydantic_model.py
```diff
@@ -318,6 +324,11 @@ def _validate_model(self) -> typing_extensions.Self:
           '`imas_uri` or `imas_filepath` to be a valid input. Inputs provided: '
           f'{specified_inputs}.'
       )
+    if self.slice_time is not None and self.slice_index != 0:
```
Nush395: why can both of these not be provided? The docstring says that time overrides index so I would assume that a user could provide both.

torax/_src/geometry/tests/imas_test.py
```diff
@@ -97,6 +98,61 @@ def test_IMAS_input_with_equilibrium_object(self):
     )
     config.build_geometry()
 
+  def test_IMAS_loading_specific_slice(self):
+    filename = 'ITERhybrid_rampup_11_time_slices_COCOS17_IDS_ddv4.nc'
+    SLICE_TIME = 40
```
Nush395: maybe just pass these in directly if they only have one usage

torax/_src/geometry/tests/imas_test.py
```diff
@@ -97,6 +98,61 @@ def test_IMAS_input_with_equilibrium_object(self):
...
+            geo_at_slice_from_time.__dict__[key],
+            geo_at_slice_from_index.__dict__[key],
+            err_msg=f'Value {key} mismatched between slice_time and slice_index',
+        )
+        with self.assertRaises(AssertionError):
+          np.testing.assert_allclose(
+              geo_at_0.__dict__[key],
+              geo_at_slice_from_time.__dict__[key],
+          )
+        with self.assertRaises(AssertionError):
```
Nush395: if you've checked the two sliced geo are equal then it's not necessary to check they both don't agree with geo at t=0

torax/_src/geometry/imas.py
```diff
@@ -72,9 +77,27 @@ def geometry_from_IMAS(
...
+    slice_index = (
+        np.searchsorted(equilibrium.time, slice_time, side="right") - 1
+    )
+    if not np.allclose(equilibrium.time[slice_index], slice_time, atol=1e-3):
+      logging.warn(
+          f"Requested t={slice_time} not in IDS; "
+          f"using t={equilibrium.time[slice_index]})"
+      )
+
+  if slice_index > len(equilibrium.time_slice):
```
Nush395: I think this should be >= as we are in python 0-indexed world

torax/data/third_party/geo/README
```diff
@@ -27,4 +27,11 @@ The mapping of datafile: LICENSE for the geometry files in this directory are as
 All associated LICENSE files are also located in this directory.
 
 ITER_hybrid_citrin_equil_cheasedata.mat2cols: LICENSE_PINT
-eqdsk_cocos02.eqdsk: LICENSE_PINT
\ No newline at end of file
+eqdsk_cocos02.eqdsk: LICENSE_PINT
+
+The IMAS netCDF file ITERhybrid_rampup_11_time_slices_COCOS17_IDS_ddv4.nc was
+generated with CHEASE using p' and ff' from a TORAX simulation using
+iterhybrid_rampup.py example config. It contains equuilibrium for each of the
```
Nush395: typo: equuilibrium -> equilibrium. also should be pluralised potentially?

torax/data/third_party/geo/README
```diff
@@ -27,4 +27,11 @@ The mapping of datafile: LICENSE for the geometry files in this directory are as
 All associated LICENSE files are also located in this directory.
 
 ITER_hybrid_citrin_equil_cheasedata.mat2cols: LICENSE_PINT
-eqdsk_cocos02.eqdsk: LICENSE_PINT
\ No newline at end of file
+eqdsk_cocos02.eqdsk: LICENSE_PINT
+
+The IMAS netCDF file ITERhybrid_rampup_11_time_slices_COCOS17_IDS_ddv4.nc was
+generated with CHEASE using p' and ff' from a TORAX simulation using
```
Nush395: "using the iterhybrid_rampup.py example config"

torax/data/third_party/geo/README
```diff
@@ -27,4 +27,11 @@ The mapping of datafile: LICENSE for the geometry files in this directory are as
...
+iterhybrid_rampup.py example config. It contains equuilibrium for each of the
+time slices of the TORAX simulation. The plasma boundaries outline R and Z
```
Nush395: "The plasma boundaries outline R and Z were prescribed constant for all time slices"

this also doesn't quite read correctly to me, could be my lack of physics knowledge though!

torax/_src/geometry/imas.py
```diff
@@ -72,9 +77,27 @@ def geometry_from_IMAS(
...
     )
-  # TODO(b/431977390): Currently only the first time slice is used, extend to
+  # TODO(b/431977390): Currently only a single time slice is used, extend to
   # support multiple time slices.
-  IMAS_data = equilibrium.time_slice[0]
+  # Convert time to index
+  if slice_time is not None:
+    # Find the closest time in the IDS that is <= slice_time
+    slice_index = (
+        np.searchsorted(equilibrium.time, slice_time, side="right") - 1
```
Nush395: is `equilibrium.time` guaranteed to be sorted?

torax/_src/geometry/imas.py
```diff
@@ -72,9 +77,27 @@ def geometry_from_IMAS(
...
+    )
+    if not np.allclose(equilibrium.time[slice_index], slice_time, atol=1e-3):
```
Nush395: I would expect the atol to be very tight here to get an exact match, why was 1e-3 chosen?

Nush395 (commented):
thanks Theo! Left a few comments

jcitrin:
@theo-brown , sorry this one slipped through the cracks.

If it's still relevant can you respond to any comments, sync, and we'll re-review?

torax/_src/geometry/imas.py
theo-brown: Unsure. Happy to add a check and then sort if it's not.

torax/_src/geometry/imas.py
theo-brown: Good point, changed to be ns tolerance

theo-brown:
No problem, it also slipped off my radar! I didn't see @Nush395 had reviewed it. Changes made! I wasn't sure whether it's ok to assume time is sorted in the equilibrium IDS, I'm not sure whether this is enforced. Can easily add a check and do a sort-then-search if you think it's required.

torax/_src/geometry/standard_geometry.py
```diff
@@ -1029,6 +1031,9 @@ def from_IMAS(
         values in the Geometry are rescaled to match the new Ip.
       n_rho: Radial grid points (num cells).
       hires_factor: High resolution factor for calculations.
+      slice_time: Time of slice to load from IMAS IDS. If given, overrides
+        slice_index.
+      slice_index: Index of slice to load from IMAS IDS (default 0).
```
Nush395: no need to say (default 0) here as that's clear from the signature!

torax/_src/imas_tools/input/equilibrium.py
```diff
@@ -46,6 +48,9 @@ def geometry_from_IMAS(
     n_rho: Radial grid points (num cells)
     hires_factor: Grid refinement factor for poloidal flux <--> plasma current
       calculations.
+    slice_time: Time of slice to load from IMAS IDS. If given, overrides
+      slice_index.
+    slice_index: Index of slice to load from IMAS IDS (default 0).
```
Nush395: similar comment about (default 0) in docstring

torax/_src/imas_tools/input/tests/equilibrium_test.py
```diff
@@ -97,6 +98,68 @@ def test_IMAS_input_with_equilibrium_object(self):
...
+        imas_filepath=filename, slice_index=5
+    )
+
+    geo_at_0 = config_at_0.build_geometry()
+    geo_at_slice_from_time = config_at_slice_from_time.build_geometry()
+    geo_at_slice_from_index = config_at_slice_from_index.build_geometry()
+
+    # Check that the geometry is not the same as at t=0
+    with self.assertRaises(AssertionError):
+      for key in geo_at_slice_from_time.__dict__.keys():
```
Nush395: maybe: worth pulling this logic into a private helper function as repeated below? Leave it up to you to decide which is clearer!

torax/data/third_party/geo/README
```diff
@@ -27,4 +27,10 @@ The mapping of datafile: LICENSE for the geometry files in this directory are as
 All associated LICENSE files are also located in this directory.
 
 ITER_hybrid_citrin_equil_cheasedata.mat2cols: LICENSE_PINT
-eqdsk_cocos02.eqdsk: LICENSE_PINT
\ No newline at end of file
+eqdsk_cocos02.eqdsk: LICENSE_PINT
+
+The IMAS netCDF file TERhybrid_rampup_11_time_slices_COCOS17_IDS_ddv4.nc was generated
```
Nush395: typo: ITER

theo-brown:
Thanks for the review @Nush395!

Changes:
- Squashed commits
- Added check if time is sorted, and sort if not
- Removed "(default 0)"
- Fixed typo in README
- Switched to assertRaisesRegex to check the error properly

torax/_src/imas_tools/input/equilibrium.py
```diff
@@ -72,9 +77,36 @@ def geometry_from_IMAS(
...
+          f"using t={equilibrium.time[slice_index]})"
+      )
+
+  if slice_index >= len(equilibrium.time_slice):
+    raise IndexError(
+        f"slice_index={slice_index} out of range for IDS with "
+        f"{len(equilibrium.time_slice)} time slices"
+    )
+  IMAS_data = equilibrium.time_slice[slice_index]
+  B_0 = np.abs(IMAS_data.global_quantities.magnetic_axis.b_field_phi)
```
Nush395, line 109: I just noticed this additional unused `B_0`, should this be removed? If so I can do so.

theo-brown: Yeah this is definitely wrong, not sure why I added it :thinking: please remove!

theo-brown:
@Nush395 I think something might've gone wrong in the copybara import.

From the merge commit: https://github.com/google-deepmind/torax/commit/2e5f70e1e8b8da810f47ed85aaa762bda7a9ceb0
https://github.com/google-deepmind/torax/blob/2e5f70e1e8b8da810f47ed85aaa762bda7a9ceb0/torax/_src/geometry/pydantic_model.py#L314-L314

From the branch: https://github.com/google-deepmind/torax/commit/5d2093b22e9b77701f7c8e48254c133d696a6e96
https://github.com/google-deepmind/torax/blob/5d2093b22e9b77701f7c8e48254c133d696a6e96/torax/_src/geometry/pydantic_model.py#L314-L314

Nush395:
> @Nush395 I think something might've gone wrong in the copybara import.
> 
> From the merge commit: [2e5f70e](https://github.com/google-deepmind/torax/commit/2e5f70e1e8b8da810f47ed85aaa762bda7a9ceb0)
> 
> https://github.com/google-deepmind/torax/blob/2e5f70e1e8b8da810f47ed85aaa762bda7a9ceb0/torax/_src/geometry/pydantic_model.py#L314-L314
> 
> From the branch: [5d2093b](https://github.com/google-deepmind/torax/commit/5d2093b22e9b77701f7c8e48254c133d696a6e96)
> 
> https://github.com/google-deepmind/torax/blob/5d2093b22e9b77701f7c8e48254c133d696a6e96/torax/_src/geometry/pydantic_model.py#L314-L314

@jcitrin commented in internal review that this `slice_time` could be negative so I modified this before submitting. Will call out any changes more explicitly in future!

jcitrin: "time can be negative, it's a user definition. There are use-cases where t=0 is some "important reference time" in the middle of a simulation"

theo-brown:
Ok, cool! Thanks for clarifying :)

PR #1436, opened 2025-08-12
Add TGLFNNukaeaTransportModel

theo-brown:
Implementing TGLF-NN-UKAEA in TORAX.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -933,6 +933,9 @@ transport model.
 * ``'qlknn'``
   A QuaLiKiz Neural Network surrogate model, the default is
   `QLKNN_7_11 <https://github.com/google-deepmind/fusion_surrogates>`_.
+* ``'tglfnn-ukaea'``
+  A TGLF Neural Network surrogate model developed by UKAEA,
+  `reimplemented in JAX <https://github.com/google-deepmind/fusion_surrogates>`_.
 * ``'qualikiz'``
   The `QuaLiKiz <https://gitlab.com/qualikiz-group/QuaLiKiz>`_ quasilinear
   gyrokinetic transport model.
@@ -1194,6 +1197,25 @@ It is recommended to not set ``qlknn_model_name``,  or
 ``An_min`` (float [default = 0.05])
   :math:`|R/L_{ne}|` value below which :math:`V_{eff}` is used instead of
   :math:`D_{eff}`, if ``DV_effective==True``.
+
+tglfnn-ukaea
+^^^^^^^^^^^^
+
+Runtime parameters for the TGLFNN-UKAEA model. If you use this model, please cite
+`L. Zanisi et al. (2025) <https://conferences.iaea.org/event/392/contributions/36059/>`_.
+
+``machine`` (str [default = 'multimachine'])
+  Machine type for TGLFNN-UKAEA. Options are 'multimachine', and 'step'.
+  The 'multimachine' model is trained on a hypercube covering a wide variety
+  of conventional tokamaks. The 'step' model is trained specifically on the
+  design space of UKAEA's Spherical Tokamak for Energy Production (STEP).
+  See `ukaea/tglfnn-ukaea <https://github.com/ukaea/tglfnn-ukaea>`_ for more
+  details. Note: currently, only the 'multimachine' model is publicly available.
+
+``An_min`` (float [default = 0.05])
+  :math:`|R/L_{ne}|` value below which :math:`V_{eff}` is used instead of
+  :math:`D_{eff}`.
+
 
 qualikiz
 ^^^^^^^^


theo-brown:
## Initial tests, TGLFNN (solid) vs QLKNN (dashed)

- All tests run by modifying `test_iterhybrid_predictor_corrector.py`.
- Note TGLFNNukaea training data may not cover ITER domains, so take with a pinch of salt.
- In TGLFNNukaea, `VEXB_SHEAR` input set to 0 - I wasn't sure what to do with this, as I believe the expression for ExB shear either needs an E-field or a rotation rate, neither of which are present in TORAX.
- TGLFNNukaea solid lines, QLKNN dashed lines.

### Te only
tglfnn-vs-qlknn-Te

### Te, Ti
tglfnn-vs-qlknn-Te-Ti

### Te, Ti, psi
tglfnn-vs-qlknn-Te-Ti-psi

### Te, Ti, psi, ne
tglfnn-vs-qlknn-Te-Ti-psi-ne

### Conclusions
- Something wrong in calculation of particle diffusivities and convectivities from fluxes? 
- Thermal diffusivities seem pretty good, behaviour broadly matching QLKNN

jcitrin:
> In TGLFNNukaea, VEXB_SHEAR input set to 0 - I wasn't sure what to do with this, as I believe the expression for ExB shear either needs an E-field or a rotation rate, neither of which are present in TORAX.

This is fine for now. Rotation in TORAX is coming this Q3.

> TGLFNNukaea solid lines, QLKNN dashed lines.

1. No need for Te only. It doesn't add much beyond Te+Ti

2. I think that the TGLF "transport barrier" at rho<0.4 (especially for Te+Ti+psi) is due to the negative magnetic shear. In QLKNN we reduce that impact with the `avoid_big_negative_s` flag which is True by default. I think this exasperates the difference. From comparisons to higher fidelity modelling we know that slab modes are underrepresented by QuaLiKiz which should still play a role at low/negative magnetic shear, hence this choice which is also in QuaLiKiz itself. So either we make something similar for TGLF and TGLFNN (if it's also relevant for TGLF), or we turn that flag off in QLKNN for the comparisons, or just do the comparison for a case without negative magnetic shear.

3. Particle transport. OMG. Yeah that doesn't look right

theo-brown:
We found an issue with the version of the NN we were using. Things are making a bit more sense now, I believe. chi profiles looking broadly good. Will look at De, Ve / particle fluxes when I next have a chance - as the NN outputs main ion flux I think there just needs to be some reweighting to get the electron flux.

theo-brown:
iterhybrid_predictor_corrector, with multimachine TGLFNNukaea vs QLKNN.

Getting reasonable matches for χₑ, χᵢ in ρ > 0.5.

Mismatch for 0.2 < ρ < 0.5 looks like difference in turbulent modes.

Mismatch for ρ < 0.2 not particularly relevant due to patch transport?

Any thoughts @lorenzozanisi @fcasson @jcitrin?


t=0
Figure_1

t=5
Figure_1

theo-brown:
Still chipping away at the output normalisation. Plots here are from a JET-like case.

I'm trying to first do it from scratch, rather than reusing the quasilinear model structures, because I've been finding it quite difficult to identify the source of the problems.

From [TGLF docs](https://gacode.io/cgyro.html#id10), for species $s$ TGLF outputs $\rm{efs} = \bar Q_s / Q_{\rm GB}$ with $Q_{\rm GB} = n_e c_s T_e (\rho_{s,{\rm unit}}/a)^2$ and $\rm{pfs} =\bar \Gamma_s / \Gamma_{\rm GB}$ with $\Gamma_{\rm GB} = n_e c_s (\rho_{s,{\rm unit}}/a)^2$. Note these normalising terms $Q_{\rm GB}$ and  $\Gamma_{\rm GB}$ are defined wrt the electron density and temperature, even if the species under consideration is ions.

I've attempted to implement this directly as follows:
```python
B_unit = (
    core_profiles.q_face
    / (2 * jnp.pi * geo.r_mid_face)  # Note: psi_TGLF is psi_TORAX/2π
    * jnp.gradient(core_profiles.psi.face_value(), geo.r_mid_face)
)
m_ref = constants.ION_PROPERTIES_DICT["D"].A * constants.CONSTANTS.mp
c_s = (
    (core_profiles.T_e.face_value() * constants.CONSTANTS.keV2J)  # [J]
    / m_ref  # [kg]
) ** 0.5
rho_s = m_ref * c_s / (constants.CONSTANTS.qe * B_unit)

Q_GB = (
    core_profiles.n_e.face_value()  # [m^-3]
    * (core_profiles.T_e.face_value() * constants.CONSTANTS.keV2J)  # [J]
    * c_s  # [ms^-1]
    * (rho_s / geo.a_minor) ** 2  # []
)
Gamma_GB = core_profiles.n_e.face_value() * c_s * (rho_s / geo.a_minor) ** 2
```
image

From dimensional analysis, I believe $Q_{\rm GB}$ is in $\rm [Wm^{-2}]$, which I think makes sense.

From the TORAX paper, 
image

If I understand correctly, the bit in the square brackets on the RHS is the flux. 
Ignoring the convective part (?), we have 

$$
Q_e = \chi_e n_e \frac{g_1}{V'} \frac{\partial T_e}{\partial \hat{\rho}}
$$

which gives

$$
\chi_e = Q_e / (n_e \frac{g_1}{V'} \frac{\partial T_e}{\partial \hat{\rho}})
$$
image

This is at least 10x smaller than expected. Do I also need to do something with the $V'$ term? Is there some geometry conversion that I'm missing?

theo-brown:
Progress update: been working on a toy case with @fcasson, we've now fixed various problems with the output normalisations and conversion to chis.

Next step is to make sure we're doing the right thing with particle transport.

Toy tokamak, initial condition, comparison between JETTO and TORAX
image

theo-brown:
Comparison with JETTO, toy case, including particle transport and pellet source:

Initial condition: shows exact match in chi profiles
image

Final condition: shows close match in T, n. Mismatch in chi_e is most likely due to solver differences (e.g. numerical  temporal smoothing/damping due to theta method)
image

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "jaxlib>=0.7.0",
     "jaxopt>=0.8.2",
     "flax>=0.10.0",
-    "fusion_surrogates==0.1.0",
+    "fusion_surrogates[tglfnnukaea]",
```
theo-brown: I tried to switch between installing `fusion_surrogates[tglfnnukaea]` and `fusion_surrogates==0.1.0` based on whether you were installing `torax[tglfnnukaea]` but I couldn't get it to work with CI.

torax/_src/transport_model/tglf_based_transport_model.py
```diff
@@ -332,4 +338,62 @@ def _prepare_tglf_inputs(
...
+    # Convert from power to chi
+    # Note: g1/vpr = ⟨(∇ρₙ)²⟩ ∂V/∂ρₙ, and has units [m]
+    dT_e_drhon = core_profiles.T_e.face_grad() * constants.CONSTANTS.keV_to_J
+    dT_i_drhon = core_profiles.T_i.face_grad() * constants.CONSTANTS.keV_to_J
+    chi_e = -P_e / (
+        core_profiles.n_e.face_value() * dT_e_drhon * geo.g1_over_vpr_face
+    )
+    chi_i = -P_i / (
+        core_profiles.n_i.face_value() * dT_i_drhon * geo.g1_over_vpr_face
+    )
```
theo-brown, line 376: The denormalisation of TGLF is slightly different to QuaLiKiz.

In particular, TGLF normalises everything with respect to the electron temperature and density, which means that chi_i can't be denormalised in the same way.

Due to the difficulties I had trying to track down bugs in this interface, I would strongly advocate for this sort of layout - denormalisation, conversion to power, conversion to chi - rather than what is done in QLKNN. 

However, if there are any speed improvements that can be made by changing what is computed, they definitely take priority.

torax/_src/transport_model/tglf_based_transport_model.py
```diff
@@ -332,4 +338,62 @@ def _prepare_tglf_inputs(
...
+    # method. This sets purely diffusive transport in regions where the flux is
+    # with the temperature gradient, otherwise it sets purely convective transport.
+    D_eff = -S_e / (core_profiles.n_e.face_grad() * geo.g1_over_vpr_face)
+    V_eff = S_e / (core_profiles.n_e.face_value() * geo.g0_face)
+    D_eff_mask = ((S_e >= 0) & (tglf_inputs.lref_over_lne >= 0)) | (
+        (S_e < 0) & (tglf_inputs.lref_over_lne < 0)
+    )
+    # For stability, we also set purely diffusive transport at some minimum
+    # threshold of the temperature gradient
+    D_eff_mask &= abs(tglf_inputs.lref_over_lne) >= transport.An_min
```
theo-brown, line 388: The D/V splitting could plausibly be reused from QLK, I just found it easier writing it out when debugging.

torax/_src/transport_model/tglfnn_ukaea_transport_model.py
```diff
@@ -0,0 +1,173 @@
...
+
+    self.model = tglfnn_ukaea_model.TGLFNNukaeaModel(
+        config=tglfnn_ukaea_config.TGLFNNukaeaModelConfig.load(
+            machine=self.machine, config_path=config_path
+        ),
+        stats=tglfnn_ukaea_config.TGLFNNukaeaModelStats.load(
+            machine=self.machine, stats_path=stats_path
+        ),
+    )
+    # TODO: memoization?
```
theo-brown: I know that QLKNN does some memoization stuff to improve load times. Is this something we could reproduce here?

torax/tests/test_data/test_iterhybrid_predictor_corrector_tglfnn_ukaea.py
```diff
@@ -0,0 +1,63 @@
...
+"""Identical to test_iterhybrid_predictor_corrector but with TGLFNNukaea
+transport"""
+from copy import deepcopy
+from pathlib import Path
+
+from torax.tests.test_data import test_iterhybrid_predictor_corrector
+
+base_path = Path(
+    '/home/theo/documents/ukaea/tglfnn-ukaea/MultiMachineHyper_1Aug25'
+)
```
theo-brown: What is the best way of loading the weights in the CI? The repo is https://github.com/ukaea/tglfnn-ukaea.

theo-brown (commented):
Summarised some open questions / details. @jcitrin this is now ready for review :)

Thanks @fcasson for all your in-depth support in getting this over the line!

Some of these questions (eg memoisation and weights loading, which is currently the cause of the failing test) may also be worth revisiting with @hamelphi on fusion-surrogates. Currently fusion_surrogates[tglfnnukaea] requires installing PyTorch (!) because our weights are distributed as .pt or .onnx, and at the start of this process it was deemed preferable to reimplement the network in JAX (which means loading weights from Pytorch .pt) rather than rely on ONNX loaders that may not have a reliable maintenance pathway.

torax/tests/test_data/test_iterhybrid_predictor_corrector_tglfnn_ukaea.nc
theo-brown, line 1: Comparison with the base test

Figure_1

torax/_src/transport_model/tglfnn_ukaea_transport_model.py
hamelphi: I think the initial reason for the memoization instead of loading the model in the constructor fro QLKNN was due to some problems with jax compilation at the time. This may have changed since it was implemented. If your current implementation works fine in an end-to-end test, I don't think memoization is necessary.

On the other hand, it may still be helpful for cases where we build many transport models within the same process. For instance, a long running TORAX process that runs many simulations in sequence. Though maybe the memoization could be implemented in fusion_surrogates rather than in TORAX.

torax/tests/test_data/test_iterhybrid_predictor_corrector_tglfnn_ukaea.py
hamelphi: We have had this kind of issues with QLKNN10D. That was one of the incentives to include the weights with the python library for qlknn_7_11. 

I believe that for non-python packages, then there needs to be an extra manual step to install the files. I would try to add it in the pytest configs After the initial checkout. https://github.com/google-deepmind/torax/blob/tglfnn-ukaea/.github/workflows/pytest.yml. Something like:
```
      - name: Checkout tglfnn-ukaea repository
        uses: actions/checkout@v4
        with:
          repository: ukaea/tglfnn-ukaea
          path: external_dependencies/tglfnn-ukaea
```

Or, if that doesn't work:
```
    - run: git clone https://github.com/ukaea/tglfnn-ukaea
```
Then the weights would probably be under `$PWD/tglfnn-ukaea`

Then you need to find a way to supply the path in the python test. This can be done using environment variables: e.g.:
```
    - run: echo "TGLFNN_PATH=$PWD/tglfnn-ukaea" >> "$GITHUB_ENV"
```

Another option would be to add a small python layer to the tglfnn-ukaea repo. I am thinking of just a simple library that would simply include the files and point to the correct path (like we do for qlknn_7_11 in fusion_surrogates), and allow installation through setuptools. This would avoid the nastiness of dealing with env variables and make everything more self-contained in Python.

theo-brown: Thanks for the pointers! I've added test workflows to the CI here and in https://github.com/google-deepmind/fusion_surrogates/pull/29 based on an env variable. If the env variable is not set, the integration test is skipped.

If you have internal test CI there might be some additional changes required, but I can't really help with that.

torax/_src/transport_model/tglfnn_ukaea_transport_model.py
theo-brown: OK, maybe this is something to leave for the time being but revisit in future. I've updated the comment to reflect this.

torax/tests/test_data/test_iterhybrid_predictor_corrector_tglfnn_ukaea.nc
jcitrin: Yes that looks roughly as expected.

In general we see more impact of trapped electron drive in TGLF compared to QuaLiKiz. In lower-density reactor-relevant scenarios like the hybrid scenario, this will be more prevalent.

Intuitively what this means is while ITG dominated (see chi_i/chi_e>1) , the increased trapped electron drive should increase the density peaking as we go closer to the ITG-TEM boundary. This is indeed what is seen in TGLF.

The increased density peaking should decrease a bit the ITG turbulence threshold, resulting in the lower Ti. The higher density due to peaking should result in more Ti-Te coupling, reducing Te/Ti, which is also what we see.

pyproject.toml
```diff
@@ -22,7 +22,8 @@ dependencies = [
     "jaxlib>=0.7.0",
     "jaxopt>=0.8.2",
     "flax>=0.10.0",
-    "fusion_surrogates==0.1.0",
+    "torch>=2.7.1", # Required for numpy>2 compatibility. See https://github.com/google-deepmind/fusion_surrogates/pull/29
```
jcitrin: If possible to store the model weights in a native JAX format in fusion_surrogates and avoid the torch dependency that would be great :)

torax/tests/test_data/test_iterhybrid_predictor_corrector_tglfnn_ukaea.py
```diff
@@ -0,0 +1,64 @@
...
+transport"""
+from copy import deepcopy
+import os
+from pathlib import Path
+
+from torax.tests.test_data import test_iterhybrid_predictor_corrector
+
+tglfnn_base_path = (
+    Path(os.getenv('TGLFNN_UKAEA_DIR', None)) / 'MultiMachineHyper_1Aug25'
+)
```
jcitrin: Could we put the open source UKAEA repo on PyPi and have it as a pip installable dependency? Then we can move away from env variables for setting model locations

jcitrin (approved):
Amazing stuff. Really excited to see this in action as a workhorse transport model in TORAX. The verification work against JETTO looks really good.

Pre-approving subject to the library dependency comments made

fcasson:
I don't want to delay this merge, but one caveat to be aware of before you use this in anger:  ExB shear is not implemented in the interface.  This will matter for JET and ITER cases, but maybe not SPARC, CMOD, STEP etc,

If there is a method for computing this in QLKNN interfaces, likely it can reused / adapted to TGLF units.

jcitrin:
> ExB shear is not implemented in the interface. This will matter for JET and ITER cases, but maybe not SPARC, CMOD, STEP etc,

Prescribed toroidal rotation profiles, neoclassical v_poloidal, and grad(Pi) terms for E_r calculation are all on the short-term roadmap for calculating ExB shear. Hopefully on order weeks.

pyproject.toml
hamelphi: The version should be upgraded to 0.4.0 to include the TGLFNN-UKAEA model. I will prepare a new release with the latest changes that should be 0.4.2

.github/workflows/pytest.yml
```diff
@@ -53,31 +53,40 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v4
+      # Checkout torax
+      - uses: actions/checkout@v4
 
-    # Install deps
-    - uses: actions/setup-python@v5
-      with:
-        python-version: ${{ matrix.python-version }}
+      # Checkout tglfnn-ukaea
```
hamelphi: Checking out the repo is causing errors in fusion_surrogates due to exceeded LFS budget.

I don't understand exactly how LFS budget works and am not sure if it is easy to increase it on our side, as I expect it depends on a Google-wide policy.

If there is a true cost associated with checking out the weights (the repo is 16Mb, not huge, but still not insignificant), then maybe we should avoid the automatic integration test.

fcasson:
From github docs looks to me like it's the ukaea budget that is used up?

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hamelphi:
> From github docs looks to me like it's the ukaea budget that is used up?

Ah yeah, that would make sense. Then we have to be careful of not draining UKAEA's budget with our automated tests. 

In any case, these test failures prevent our copybara workflow from merging new changes in, so I would be wary of adding the dependency to the external repo until this is resolved.

fcasson:
Can't we start the test pipeline from jax weights stored in fusion surrogates?

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hamelphi:
> Can't we start the test pipeline from jax weights stored in fusion surrogates?

There are currently no weights stored in fusion_surrogates for TGLFNN, it only contains inference code.

theo-brown:
As of this morning, proposed new method for distributing weights via pip installable package currently under review in internal UKAEA repos. This would massively simplify loading here and in fusion_surrogates.

pyproject.toml
hamelphi: 0.4.2 is ready. It might be worth trying again with the new version. I suspect the trouble you were having could have been due to incompatible versions. Something like:
```
depedencies = [
 ...
 fusion_surrogates==0.4.2
 ]
 
 ...
 [project.optional-dependencies]
 dev = [
   ...
    fusion_surrogates[tglfnnukaea]==0.4.2
    ]
  tglfnnukaea = [
    fusion_surrogates[tglfnnukaea]==0.4.2
]

pyproject.toml
hamelphi: This should be included with the fusion_surrogates dependency. I don't think we need it explicitly here.

theo-brown:
@jcitrin @hamelphi updated to reflect https://github.com/google-deepmind/fusion_surrogates/pull/32. This fixes the problems with Git LFS, and also simplifies the dependency tree. 

Tests are failing until fusion_surrogates PR is merged and new version is published to PyPI. Tests pass on a local install, with the version of fusion surrogates from  https://github.com/google-deepmind/fusion_surrogates/pull/32.
Lmk when it's in and I'll manually retrigger the tests.

pyproject.toml
theo-brown: Revisit when 0.4.3 is pushed

hamelphi: Done. 0.4.3 is ready.
I suggest making  `torax[tglfnnukaea]` a depedency of `dev` instead of repeating the fusion_surrogates dep. See https://github.com/google-deepmind/fusion_surrogates/pull/34

theo-brown: This is a nice pattern, thanks for the suggestion!

theo-brown:
Forgot about docs! Added now.

jcitrin:
Looking at the pyproject.toml: it looks like tglfnnukaea is not installed by default. Wouldn't it be easier for users to have it installed by default? It may be a workhorse model so users may be surprised if they don't have it from the getgo

hamelphi:
> Looking at the pyproject.toml: it looks like tglfnnukaea is not installed by default. Wouldn't it be easier for users to have it installed by default? It may be a workhorse model so users may be surprised if they don't have it from the getgo

Sounds good to me. We should keep in mind that this will add 60Mb to the deps. Though the current default install is around 1Gb (the largest chunk being jaxlib at 300Mb), so it is relatively small.

We could also make it default for fusion_surrogates.

theo-brown:
>  Wouldn't it be easier for users to have it installed by default? It may be a workhorse model

Yep, not averse to this!

Before we do, it might be good to do some speed/performance testing, because I've found that some change over the last few commits has caused a big drop in performance. Could we have it as an optional dependency for the time being until it's a bit more stress-tested?

jcitrin:
> Yep, not averse to this! Before we do, it might be good to do some speed/performance testing, because I've found that some change over the last few commits has caused a big drop in performance. Could we have it as an optional dependency for the time being until it's a bit more stress-tested?

So shall we keep it as is for now, and then in a future PR make it default once the performance issues are solved?

theo-brown:
Sounds good to me!

theo-brown:
Timing comparison:

image

hamelphi:
Thanks for the benchmark analysis. 

I am not surprised that QLKNN_7_11 is faster than TGLFNN since it is a much smaller model. The specific model was picked with inference time as one of the metrics in mind. That also explains why GPUs do not help in that case, since I suppose the overhead is large compared to the inference time. GPUs could become more relevant for QLKNN with large batch inference.

On the flip side, I do expect the quality of TGLFNN predictions to be more accurate, or at least to have a closer correlation with the base model.

jcitrin:
Congratulations @theo-brown , @lorenzozanisi ! It's a great milestone having this in TORAX. Looking forward to using it! Rotation will come soon.

@fcasson

lorenzozanisi:
Thanks @theo-brown @hamelphi @fcasson  and @jcitrin  for pulling this together - excited to see the NNs in action within TORAX!

PR #1439, opened 2025-08-13
Feature/muscle3 actor

mikesndrs:
Muscle3 actor for torax
Added:
- Muscle3 actor
- Optional muscle3 dependency
- Actor documentation
- muscle3 workflow test 

side:
- fixed numpy concatenate
- fixed f shift bug due to wrong B0 input
- option to skip z_magnetic_axis input

### DIFF
--- a/torax/_src/orchestration/muscle3_utils.py
+++ b/torax/_src/orchestration/muscle3_utils.py
@@ -7,16 +7,16 @@ from typing import Optional
 
 from imas import DBEntry
 from libmuscle import Instance
-from torax._src.config.build_runtime_params import DynamicRuntimeParamsSliceProvider
-from torax._src.config.build_runtime_params import get_consistent_dynamic_runtime_params_slice_and_geometry
+from torax._src.config.build_runtime_params import RuntimeParamsProvider
+from torax._src.config.build_runtime_params import get_consistent_runtime_params_and_geometry
 from torax._src.config.config_loader import build_torax_config_from_file
-from torax._src.config.runtime_params_slice import DynamicRuntimeParamsSlice
+from torax._src.config.runtime_params import RuntimeParams
 from torax._src.geometry.geometry_provider import GeometryProvider
 from torax._src.orchestration import initial_state as initial_state_lib
 from torax._src.orchestration import sim_state
 from torax._src.orchestration.run_simulation import prepare_simulation
 from torax._src.orchestration.step_function import SimulationStepFn
-from torax._src.output_tools.imas import torax_state_to_imas_equilibrium
+from torax._src.imas_tools.output.equilibrium import torax_state_to_imas_equilibrium
 from torax._src.state import SimError
 from ymmsl import Operator
 from ymmsl import SettingValue
--- a/torax/_src/orchestration/run_muscle3.py
+++ b/torax/_src/orchestration/run_muscle3.py
@@ -21,7 +21,7 @@ from libmuscle import Instance
 from libmuscle import Message
 import numpy as np
 from torax._src.config import build_runtime_params
-from torax._src.config.build_runtime_params import get_consistent_dynamic_runtime_params_slice_and_geometry
+from torax._src.config.build_runtime_params import get_consistent_runtime_params_and_geometry
 from torax._src.config.config_loader import build_torax_config_from_file
 from torax._src.geometry import geometry
 from torax._src.geometry.pydantic_model import Geometry
@@ -29,12 +29,13 @@ from torax._src.geometry.pydantic_model import GeometryConfig
 from torax._src.geometry.pydantic_model import IMASConfig
 from torax._src.orchestration import initial_state as initial_state_lib
 from torax._src.orchestration import sim_state
+from torax._src.orchestration import step_function
 from torax._src.orchestration.muscle3_utils import ExtraVarCollection
 from torax._src.orchestration.muscle3_utils import get_geometry_config_dict
 from torax._src.orchestration.muscle3_utils import get_setting_optional
 from torax._src.orchestration.muscle3_utils import merge_extra_vars
 from torax._src.orchestration.run_simulation import prepare_simulation
-from torax._src.output_tools.imas import torax_state_to_imas_equilibrium
+from torax._src.imas_tools.output.equilibrium import torax_state_to_imas_equilibrium
 from torax._src.state import SimError
 from ymmsl import Operator
 
@@ -47,7 +48,7 @@ class ToraxMuscleRunner:
     output_all_timeslices: bool = False
     db_out: Optional[IDSToplevel] = None
     torax_config = None
-    dynamic_runtime_params_slice_provider = None
+    runtime_params_provider = None
     step_fn = None
     time_step_calculator_dynamic_params = None
     equilibrium_interval = None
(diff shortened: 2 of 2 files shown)


torax/_src/orchestration/tests/muscle_actor_test.py
```diff
@@ -0,0 +1,123 @@
...
+ymmsl = pytest.importorskip("ymmsl")
+
+from torax._src.orchestration.run_muscle3 import main as torax_actor
+
+# imas_core is required for IDS serialize, unfortunately this means we cannot run these
+# tests in github Actions yet..
+pytest.importorskip("imas_core")
+
+
+def source_for_tests():
```
jcitrin, line 16: micronit: maybe rename to "data_source" and "data_sink" or "data_target" , since in integrated modelling "source" and "sink" have clear meanings for terms in the transport PDEs, and it's confusing to see them used here for equilibrium data sources and data targets

torax/_src/orchestration/run_muscle3.py
```diff
@@ -0,0 +1,184 @@
...
+from torax._src.config import build_runtime_params
+from torax._src.config.build_runtime_params import get_consistent_dynamic_runtime_params_slice_and_geometry
+from torax._src.config.config_loader import build_torax_config_from_file
+from torax._src.orchestration import initial_state as initial_state_lib
+from torax._src.orchestration.muscle3_utils import get_geometry_config_dict
+from torax._src.orchestration.muscle3_utils import get_setting_optional
+from torax._src.orchestration.muscle3_utils import get_t_next
+from torax._src.orchestration.muscle3_utils import merge_extra_vars
+from torax._src.orchestration.muscle3_utils import receive_equilibrium
+from torax._src.orchestration.muscle3_utils import receive_ids_through_muscle3
```
jcitrin: Can remove? It looks like this is only used within muscle3_utils as a helper method for receive_equilibrium. Can then make it private in muscle3_utils

torax/_src/orchestration/muscle3_utils.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def receive_equilibrium(
+    instance: Instance,
+    geometry_provider: GeometryProvider,
+    state: sim_state.ToraxSimState,
+    config: model_config.ToraxConfig,
+    port_name: str,
+) -> Tuple[GeometryProvider, float, Optional[float], ExtraVarCollection]:
+    # TODO: receive all inputs [equilibrium, core_profiles, core_profiles]
```
jcitrin: typo, core_sources?

torax/_src/orchestration/muscle3_utils.py
```diff
@@ -0,0 +1,230 @@
...
+        return geometry_provider, state.t, None, ExtraVarCollection()
+    equilibrium_data, t_cur, t_next = receive_ids_through_muscle3(
+        instance, port_name, 'equilibrium'
+    )
+
+    geometry_configs = {}
+    extra_var_col = ExtraVarCollection()
+    torax_config_dict = get_geometry_config_dict(config)
+    torax_config_dict["geometry_type"] = "imas"
+    with DBEntry("imas:memory?path=/", "w") as db:
```
jcitrin: Could use some comments here. As I understand it, this provides a geometry provider with all the time slices existent in the sent equilibrium IDS?

torax/_src/orchestration/run_muscle3.py
```diff
@@ -0,0 +1,184 @@
...
+        # F_INIT
+        if instance.is_connected("equilibrium_f_init"):
+            step_fn._geometry_provider, t_cur, t_next_outer, extra_var_col = receive_equilibrium(
+                instance,
+                step_fn._geometry_provider,
+                sim_state,
+                torax_config,
+                "equilibrium_f_init",
+            )
+        if first_run or instance.is_connected("equilibrium_f_init"):
```
jcitrin: I think this can be cleaner with respect to the branches "equilibrium_f_init_ exists" vs "not".

There are duplicated calculations. `prepare_simulation` should only be done when we have the "correct" initial config, which depends on whether `equilibrium_f_init` is connected or not. By reordering things, I think you can avoid the recalculation of specific prepare_simulation outputs.

torax/_src/orchestration/run_muscle3.py
```diff
@@ -0,0 +1,184 @@
...
+                    t=torax_config.numerics.t_initial,
+                    static_runtime_params_slice=static_runtime_params_slice,
+                    dynamic_runtime_params_slice_provider=dynamic_runtime_params_slice_provider,
+                    geometry_provider=step_fn._geometry_provider,
+                    step_fn=step_fn,
+                )
+            )
+        first_run = False
+
+        equilibrium_data = torax_state_to_imas_equilibrium(sim_state, post_processed_outputs)
```
jcitrin: nit: both here and below, can put this (and merge_extra_vars) within the `output_all_timeslices` if statement, to avoid unnecessary computation if not needed

torax/_src/orchestration/run_muscle3.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+            send_ids_through_muscle3(
+                instance,
+                equilibrium_data,
+                "equilibrium_o_i",
+                sim_state.t,
+                t_next=t_next_inner,
+            )
+
+            # S
```
jcitrin: Some more comments would be beneficial here clarifying the role of equilibrium_init vs equilibrium_s.

For equilibrium_init, I can imagine that in some workflows, then incoming equilibrium would even be the entire time-series. 

However, for equilibrium_s,  which is inter-step, then it only makes sense for a single equilibrium slice to be included in the provider, which will then be used for both geo_t and geo_t_plus_dt.

If both equilibrium_init and equilibrium_s are present, then for the first time in the loop the role of the first equilibrium_s seems duplicate to equilibrium_init, which in this case should only provide a single initial timeslice

torax/_src/orchestration/run_muscle3.py
```diff
@@ -0,0 +1,184 @@
...
+            if instance.is_connected("equilibrium_s"):
+                step_fn._geometry_provider, t_cur, _, extra_var_col = receive_equilibrium(
+                    instance,
+                    step_fn._geometry_provider,
+                    sim_state,
+                    torax_config,
+                    "equilibrium_s",
+                )
```
jcitrin, line 228: Another note on equilibrium_s, and the need to modify the internal TORAX API for such "new" cases like this where new equilibria are provided inter_step.

Inside the step function, we have _get_geo_and_dynamic_runtime_params_at_t_plus_dt_and_phibdot . This uses geo_t and geo_t_plus_dt to calculate the phibdot term, which is non-zero is geo_t and geo_t_plus_dt have specific differences (e.g. if the toroidal flux at the LCFS is different). This can be important in rampup and rampdown.

However, in workflows with equilibrium_s, I assume that a single new timeslice is provided to then be used inside step function as geo_t and geo_t_plus_dt. This then means the phibdot=0 which is not accurate in general.

Maybe a TODO can be added here. There needs to be a way to treat this. There are several options that can be discussed.

mikesndrs:
TORAX-MUSCLE3 actor now available in
https://github.com/iterorganization/TORAX-MUSCLE3/
closing this PR

PR #1443, opened 2025-08-14
Implementation and testing for reading and writing IMAS core/plasma_profiles IDS

MateoBell:
Add functions to read and write core_profiles or plasma_profiles IDSs to init and save TORAX simulations. 

- [x] Documentation

Implementation

Works with DDv4.x.
- [x] Possibility to read multiple time slices in input.
- [x] Possibility to write one time slice in output
- [x] Details for ion composition, for each of the main ions and impurities in the output transfer function.
- [x] Output profiles written on cell_plus_boundaries grid (from https://github.com/google-deepmind/torax/pull/977 and https://github.com/google-deepmind/torax/pull/978)
&#8594; Done **except for j_ohmic and j_external** for which we only have cell grid available.

Testing
- [x]  Unit tests for input and output.
- [x] Add 2 netCDF core_profiles IDS in DDv4 to be able to run the tests without requiring access to IMAS-core. First one contains simple profiles based on iterhybrid_rampup config, and the second one is based on an ITER case from METIS.

To implement (possibly left for future PR):
- [ ] Read impurity fractions from core_profiles instead of dynamic_runtime_params_slice with https://github.com/google-deepmind/torax/pull/1408.
- [ ] Possibility to write multiple time slices in output.


docs/configuration.rst
```diff
@@ -1862,6 +1862,18 @@ There are three main methods to load IMAS equilibrium:
 * Providing the equilibium IDS on the fly (equilibrium_object).
   Using this method the IDS can be provided externally or pre-loaded.
 
+Loading Profiles
+^^^^^^^^^^^^^^^^
+
+The initial and optionnally prescribed time-dependence of the temperature, density and current profiles given in the config can be loaded from any IMAS core_profiles or plasma_profiles IDS.
```
jcitrin: typo: optionally

docs/configuration.rst
jcitrin, line 1: maintain 80 chars per line

docs/configuration.rst
```diff
@@ -1862,6 +1862,18 @@ There are three main methods to load IMAS equilibrium:
...
+The IDSs needs to be in Data Dictionary version 4.0.0 or newer.
+The mapping function :todo:`add the actual link to the function` |core_profiles_from_imas| returns a dictionary with the data from core_profiles to load into the config. The IDS needs to be provided on the fly as it can be done for the geometry.
```
jcitrin: do the todos here and below :)

jcitrin: what do you mean by "provided on the fly"?

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+from torax._src.orchestration import run_loop
+from torax._src.orchestration.run_simulation import prepare_simulation
+from torax._src.test_utils import sim_test_case
+from torax._src.torax_pydantic import model_config
+from torax.imas_tools.core_profiles import core_profiles_from_IMAS
+from torax.imas_tools.core_profiles import core_profiles_to_IMAS
+from torax.imas_tools.core_profiles import update_dict
+
+
+class Core_profilesTest(sim_test_case.SimTestCase):
```
jcitrin: CoreProfilesTest

torax/data/third_party/geo/core_profiles_15MA_DT_50_50_flat_top_slice.nc
jcitrin, line 1: can put these in a new folder in third_party, like "imasdata"

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+        # Input core_profiles reading and config loading
+        config = self._get_config_dict("test_iterhybrid_rampup_short.py")
+
+        path = 'core_profiles_ddv4_iterhybrid_rampup_conditions.nc'
+        core_profiles_in = _load_imas_data(path, "core_profiles")
+
+        # Modifying the input config profiles_conditions class
+        core_profiles_conditions = core_profiles_from_IMAS(core_profiles_in, read_psi_from_geo=True)
+        config_with_IMAS_profiles = update_dict(config, core_profiles_conditions) #Is it better to do like this, or first convert to ToraxConfig and use config.config_args.recursive_replace or maybe another function that does the same instead ?
+        # Or use ToraxConfig.update_fields ?
```
jcitrin: yes, using ToraxConfig.update_fields is the way to go. Look at other examples in the codebase

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def test_run_with_core_profiles_to_IMAS(
+        self,
+    ):
+        """Test that TORAX simulation example can run with input core_profiles ids profiles, without raising error."""
+
+        # Input core_profiles reading and config loading
+        config = self._get_config_dict("test_iterhybrid_rampup_short.py")
+
+        path = 'core_profiles_ddv4_iterhybrid_rampup_conditions.nc'
```
jcitrin: provide comments saying where this is from

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+        torax_config = model_config.ToraxConfig.from_dict(config_with_IMAS_profiles)
+
+        #Run Sim
+        _, results = torax.run_simulation(torax_config)
+        # Check that the simulation completed successfully.
+        if results.sim_error != torax.SimError.NO_ERROR:
+          raise ValueError(
+              f'TORAX failed to run the simulation with error: {results.sim_error}.'
+          )
+
```
jcitrin: This is a prerequisite, but not a strong enough test.

Ideally we would do a "round trip". You'd run TORAX with a regular config, take the outputs and convert them to IMAS, then modify the original config and update with the IMAS core profiles, then rerun the simulation, compare results, and see that they're the same

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+    ):
+      """Test to compare initialized profiles in TORAX with the initial core_profiles used to check consistency."""
+
+      if rtol is None:
+          rtol = self.rtol
+      if atol is None:
+          atol = self.atol
+      # Input core_profiles reading and config loading
+      config = self._get_config_dict(config_name)
+      # path = 'core_profiles_ddv4_iterhybrid_rampup_conditions.nc'
```
jcitrin: remove

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+      path = 'core_profiles_15MA_DT_50_50_flat_top_slice.nc' #Using this as input instead of rampup_conditions because it has more radial resolution.
+      core_profiles_in = _load_imas_data(path, "core_profiles")
+      rhon_in = core_profiles_in.profiles_1d[0].grid.rho_tor_norm
+
+      # Modifying the input config profiles_conditions class
+      core_profiles_conditions = core_profiles_from_IMAS(core_profiles_in, read_psi_from_geo= False)
+      config_with_IMAS_profiles = update_dict(config, core_profiles_conditions) #Is it better to do like this, or first convert to ToraxConfig and use config.config_args.recursive_replace or maybe another function that does the same instead ?
```
jcitrin: as above. Make input appropriate for ToraxConfig.update_fields and use that

torax/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,184 @@
...
+        _, results = torax.run_simulation(torax_config)
+        # Check that the simulation completed successfully.
+        if results.sim_error != torax.SimError.NO_ERROR:
+          raise ValueError(
+              f'TORAX failed to run the simulation with error: {results.sim_error}.'
+          )
+
+    @parameterized.parameters(
+        [
+            dict(config_name="test_iterhybrid_rampup_short.py", rtol=0.02, atol=1e-8),
```
jcitrin: this high rtol is a bit weird for me. These are just simple 1d profiles stored in two different data structures. The agreement should be almost perfect. What's going on?

docs/configuration.rst
jcitrin, line 1: In general, I think at this stage how to use IMAS inputs isn't yet exposed to the user (the tests do this "manually" with update fields), so can do the documentation in another followup PR. It feels like there are missing instructions of how to set the IMAS input file, to toggle "IMAS mode" on, various interactions with how impurities are set (impurity_mode should still be set by the user here), etc. Maybe there's a desire to shift the IMAS IDS times to coincide with TORAX times, similar to what we can do for equilibrium. Can discuss further.

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  profiles_1d = ids.profiles_1d
+  time_array = [float(profiles_1d[i].time) for i in range(len(profiles_1d))]
+  rhon_array = [
+      profiles_1d[i].grid.rho_tor_norm for i in range(len(profiles_1d))
+  ]
+  # numerics
+  t_initial = float(profiles_1d[0].time)
+
+  # plasma_composition
+  # Zeff taken from here or set into config before ?
```
jcitrin: This depends on the impurity_mode. For 'n_e_ratios', Z_eff should not be set in the config, it's emergent from the impurity densities. For 'n_e_ratios_Z_eff', then we do set Z_eff, but one of the impurity species values needs to be "None", and it's then found internally. Even in IMAS-mode, the user should be able to set which impurity density is "unknown". 

The "fractions" impurity mode should not be used with IMAS, it doesn't make sense. 

Can modify (in future PR) the Pydantic validations for ProfileConditions to check if we are in IMAS-mode (via a new ProfileConditions attribute?) and then validate stuff. We may also need a new attribute for the "unknown impurity" when in IMAS-mode + n_e_ratios_Zeff impurity_mode

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  Z_eff = {
+      time_array[ti]: {
+          rhon_array[ti][rj]: profiles_1d[ti].zeff[rj]
+          for rj in range(len(rhon_array[ti]))
+      }
+      for ti in range(len(time_array))
+  }
+
+  # profile_conditions
+  # Should we shift it to get psi=0 at the center ?
```
jcitrin: good question. I think for consistency it should be done the same way as in geometry, which doesn't shift I think. It doesn't matter for the internal calculations, but it may matter if we are keeping track of flux consumption across a workflow and thus we shouldn't modify the absolute value of psi

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+
+  T_e = {
+      time_array[ti]: {
+          rhon_array[ti][rj]: profiles_1d[ti].electrons.temperature[rj] / 1e3
+          for rj in range(len(rhon_array[ti]))
+      }
+      for ti in range(len(time_array))
+  }
+  # bound_right conditions commented: Will raise an error if rhon[-1]!= 1.
+  # Te_bound_right = {time_array[ti]: profiles_1d[ti].electrons.temperature[-1]/1e3 for ti in range(len(time_array))}
```
jcitrin: If we assume that the BC is always correctly in the IMAS data with rhon[-1]=1, then I guess we can just delete these lines or just keep the comment on the assumption that rhon[-1]=1?

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  if len(profiles_1d[0].t_i_average > 0):
```
jcitrin: does this really happen that t_i_average is sometimes not filled? It is quite heavy to need to catch all the edge cases and do the averaging ourselves, and not trust that the input IDS is properly filled.

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  }
+  # ne_bound_right = {time_array[ti]: profiles_1d[ti].electrons.density[-1]for ti in range(len(time_array))}
+
+  if len(
+      ids.global_quantities.v_loop > 0
+  ):  # Map v_loop_lcfs in case it is used as bc for psi equation.
+    v_loop_lcfs = {
+        time_array[ti]: ids.global_quantities.v_loop[ti]
+        for ti in range(len(time_array))
+    }  # + or -1 ?
```
jcitrin: uhhh... not sure. Maybe do a TODO. This is the kind of thing that will be exposed very quickly when using it..

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  else:
+    v_loop_lcfs = [0.0]
+
+  return {
+      "plasma_composition": {
+          "Z_eff": Z_eff,
```
jcitrin: we are missing all the impurity info. Can be done in a followup CL

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  ids.global_quantities.beta_tor.resize(1)
+  ids.global_quantities.beta_tor_norm.resize(1)
+  ids.global_quantities.t_e_volume_average.resize(1)
+  ids.global_quantities.n_e_volume_average.resize(1)
+
+  ids.profiles_1d.resize(1)
+  ids.profiles_1d[0].time = t
+
+  ids.vacuum_toroidal_field.r0 = geometry.R_major
+  ids.vacuum_toroidal_field.b0[0] = geometry.B_0  # +1 or -1 ?
```
jcitrin: Not sure. I think we need to formally set TORAX COCOS and then this will be known. Make a TODO

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  )
+  ids.global_quantities.ion_time_slice = t
+
+  ids.profiles_1d[0].grid.rho_tor_norm = np.concatenate(
+      [[0.0], geometry.torax_mesh.cell_centers, [1.0]]
+  )
+  Phi = np.concatenate(
+      [[geometry.Phi_face[0]], geometry.Phi, [geometry.Phi_face[-1]]]
+  )
+  ids.profiles_1d[0].grid.rho_tor = np.sqrt(Phi / (np.pi * geometry.B_0))
```
jcitrin: Why not directly use rho_norm from geometry?

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  ids.profiles_1d[0].electrons.temperature = (
+      cp_state.T_e.cell_plus_boundaries() * 1e3
+  )
+  ids.profiles_1d[0].electrons.density = cp_state.n_e.cell_plus_boundaries()
+  ids.profiles_1d[0].electrons.density_thermal = (
+      cp_state.n_e.cell_plus_boundaries()
+  )
+  ids.profiles_1d[0].electrons.density_fast = np.zeros(
+      len(ids.profiles_1d[0].grid.rho_tor_norm)
+  )  # TODO: check if we can deduce the density thermal and fast somehow
```
jcitrin: Not yet. We have no fast ion density in TORAX yet. It's on the roadmap

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+  ids.profiles_1d[0].t_i_average = cp_state.T_i.cell_plus_boundaries() * 1e3
+  ids.profiles_1d[0].n_i_total_over_n_e = (
+      cp_state.n_i.cell_plus_boundaries()
+      + cp_state.n_impurity.cell_plus_boundaries()
+  ) / cp_state.n_e.cell_plus_boundaries()
+  Z_eff = np.concatenate(
+      [[cp_state.Z_eff_face[0]], cp_state.Z_eff, [cp_state.Z_eff_face[-1]]]
+  )
+  ids.profiles_1d[0].zeff = (
+      Z_eff  # Keep the formula below instead or keep zeff from cp as source of truth ?
```
jcitrin: Keep Z_eff from cp as a source of truth

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+          config.plasma_composition.get_main_ion_names(),
+          dynamic_runtime_params_slice.plasma_composition.main_ion.fractions,
+      )
+  )
+  impurities = list(
+      zip(
+          config.plasma_composition.get_impurity_names(),
+          dynamic_runtime_params_slice.plasma_composition.impurity.fractions,
+      )
+  )
```
jcitrin: I'm not going to review anything to do with impurities since we have all the new impurity API stuff and this changes things. We also have impurity_fractions now in core_profiles which makes things easier.

However, due to various shenanigans related to how we define the single effective n_impurity from a bunch of impurities, to reproduce the true individual densities, we need to calculate an impurity_density_scaling. Then, n_impurity_individual = fraction_individual * impurity_density_scaling * n_impurity 

See the calculation and the comment here. I recommend making a little helper function to reproduce what's done in the link regarding `impurity_density_scaling`  https://github.com/google-deepmind/torax/blob/01c16672cae03b794b806c89d0fc4389280075a4/torax/_src/sources/impurity_radiation_heat_sink/impurity_radiation_mavrin_fit.py#L227

jcitrin: To make things a bit more complicated (sorry), the same assumptions made when defining n_impurity, means that n_i is actually the "true" total n_i, so for main ions fraction*n_i is fine for the individual densities, but you need the impurity_density_scaling for the impurities.

In the near future (order weeks?) we will move TORAX to also internally have individual ions and impurities, so this will be easier.

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+      ids.profiles_1d[0].ion[iion].label = symbol
+    ids.profiles_1d[0].ion[iion].temperature = (
+        cp_state.T_i.cell_plus_boundaries() * 1e3
+    )
+    ids.profiles_1d[0].ion[iion].density = (
+        cp_state.n_i.cell_plus_boundaries() * frac
+    )
+    ids.profiles_1d[0].ion[iion].density_thermal = (
+        cp_state.n_i.cell_plus_boundaries() * frac
+    )  # Is it true that all ions are thermal ?
```
jcitrin: yes

torax/imas_tools/core_profiles.py
```diff
@@ -0,0 +1,474 @@
...
+    )
+    ids.profiles_1d[0].ion[num_of_main_ions + iion].pressure_thermal = (
+        total_ions_mixture_fraction
+        * post_processed_outputs.pressure_thermal_i.cell_plus_boundaries()
+    )
+    ids.profiles_1d[0].ion[num_of_main_ions + iion].element.resize(1)
+    ids.profiles_1d[0].ion[num_of_main_ions + iion].element[
+        0
+    ].a = ion_properties.A
+    ids.profiles_1d[0].ion[num_of_main_ions + iion].element[
```
jcitrin: What does IMAS assume here, the total Z or the actual charge state with partial ionization (for heavier impurities)? If it's the latter then this is incorrect and since that information isn't yet in CoreProfiles, you'd need to calculate it like in the code. It's not too bad to calculate it, just calling [this function](https://github.com/google-deepmind/torax/blob/01c16672cae03b794b806c89d0fc4389280075a4/torax/_src/physics/charge_states.py#L121)

torax/_src/imas_tools/tests/core_profiles_test.py
```diff
@@ -0,0 +1,196 @@
...
+      config = self._get_config_dict("test_iterhybrid_rampup_short.py")
+      path = 'core_profiles_ddv4_iterhybrid_rampup_conditions.nc'
+      dir = os.path.join(torax.__path__[0], 'data/third_party/imas_data')
+      core_profiles_in = _load_imas_data(path, "core_profiles", geometry_directory=dir)
+
+      # Modifying the input config profiles_conditions class
+      core_profiles_conditions = core_profiles_from_IMAS(core_profiles_in, read_psi_from_geo = False)
+      config['geometry']['n_rho']=20
+      torax_config = model_config.ToraxConfig.from_dict(config)
+      torax_config.update_fields(core_profiles_conditions)
```
tamaranorman: the other option instead of doing this is to do it using dict manipulation instead of needing to do update fields

Then the dict created in imas could then match the structure of the torax config and will make it easier to visualise if the user wants to load from imas then edit things

It also likely will be slightly more efficient

torax/_src/geometry/imas.py
```diff
@@ -177,14 +179,17 @@ def geometry_from_IMAS(
       "z_magnetic_axis": z_magnetic_axis,
   }
 
-
+#TODO: For future PR, if imas files get refactored, modify the function to be
```
tamaranorman: Can this be refactored now into a function which takes a full uri and ids_name and then one which takes the geo directory, concatenates using the path join and then calls this

docs/configuration.rst
```diff
@@ -1862,6 +1861,18 @@ There are three main methods to load IMAS equilibrium:
...
 
+Loading Profiles
+^^^^^^^^^^^^^^^^
+
+The initial and optionally prescribed time-dependence of the temperature, density and current profiles given in the config can be loaded from any IMAS core_profiles or plasma_profiles IDS.
+The IDSs needs to be in Data Dictionary version 4.0.0 or newer.
+The mapping function ``core_profiles_from_imas`` in |core_profiles_imas| returns a dictionary with the data from core_profiles to load into the config. The IDS needs to be provided on the fly as it can be done for the geometry, i.e. pre-loaded in Python from an IMAS netCDF file or another method and directly fed to the function. A more user friendly API will come in a follow-up PR with updated doc.
+The IDS object can be loaded from an IMAS netCDF file using the `IMAS-Python library <https://imas-python.readthedocs.io/en/latest/netcdf.html>`_ .
+
+It is then possible to start a simulation by loading a pre-existing config and then update it with the dictionary containing the IMAS profiles using the function ``update_dict`` from |core_profiles_imas|.
```
tamaranorman: I am not sure this is the way I would recommend doing this but updating the config dict before. We should explain a bit more how these should be use

torax/data/third_party/geo/core_profiles_15MA_DT_50_50_flat_top_slice.nc
MateoBell: do you want to move all IMAS data into these new folder or keep the IMAS equilibria into third_party/geo ?

torax/imas_tools/tests/core_profiles_test.py
MateoBell: It was comparing initial IDS profiles with profiles that went back and forth to TORAX, so they were interpolated twice. Guess that was the reason for the needed high rtol for a few values, mostly around axis and edge. Now it interpolates IDS profiles to TORAX grid so the agreement is indeed almost perfect, have been able to lower rtol a lot. I think it should be okay this way?

torax/imas_tools/core_profiles.py
MateoBell: Ok thanks, then it's better not to shift I think

torax/imas_tools/core_profiles.py
MateoBell: Don't really know, I think I mostly encountered cases where it was filled, so I can remove the if to make it lighter.

torax/imas_tools/core_profiles.py
MateoBell: From the DD documentation: https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/core_profiles.html#core_profiles-profiles_1d-ion-element-z_n 
Z_n should be the total Z, the actual charge state with partial ionization should go into profiles_1d/ion/z_ion_1d I think ? I will add it to the mapping along with impurity related things.

torax/_src/imas_tools/tests/core_profiles_test.py
MateoBell: > yes, using ToraxConfig.update_fields is the way to go. Look at other examples in the codebase

@jcitrin recommended me to use update_fields instead of dict manipulation, but I can switch back to the previous version, depending on what you think is better.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+import imas
+from imas import ids_toplevel
+from imas.ids_toplevel import IDSToplevel
+import numpy as np
+import torax
+
+
+def core_profiles_from_IMAS(
+    ids: IDSToplevel,
+    read_psi_from_geo: bool = True,
```
tamaranorman, line 30: Remove this now we have the psi_source enum in the config itself and that can just be set?

Reading psi feels cheap so we may as well just always read it

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+
+"""Useful functions to load IMAS core_profiles or plasma_profiles IDSs and
+converts them into TORAX objects.
+"""
+import os
+from typing import Any
+
+import imas
+from imas import ids_toplevel
+from imas.ids_toplevel import IDSToplevel
```
tamaranorman, line 23: Within TORAX we don't import symbols following this style guide https://google.github.io/styleguide/pyguide.html#22-imports

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,385 @@
...
+  )
+  ids.profiles_1d[0].j_bootstrap = -1 * j_bootstrap
+  sigma = np.concatenate(
+      [[cp_state.sigma_face[0]], cp_state.sigma, [cp_state.sigma_face[-1]]]
+  )
+  ids.profiles_1d[0].conductivity_parallel = sigma
+  return ids
+
+
+def calculate_impurity_density_scaling_and_charge_states(
```
tamaranorman, line 347: Is it anticipated that end users will use this? If not then maybe we can switch it such that this is private

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,385 @@
...
+from torax._src.orchestration.sim_state import ToraxSimState
+from torax._src.output_tools import post_processing
+from torax._src.physics import charge_states
+from torax._src.torax_pydantic import model_config
+
+
+# TODO: Add option to save entire state history in one core_profiles output.
+def core_profiles_to_IMAS(
+    config: model_config.ToraxConfig,
+    runtime_params_slice: runtime_params_slice.RuntimeParams,
```
tamaranorman, line 43: Call this runtime_params now as the terminology has changed

torax/_src/imas_tools/output/tests/core_profiles_test.py
```diff
@@ -0,0 +1,170 @@
...
+    dir = os.path.join(torax.__path__[0], 'data/third_party/imas_data')
+    core_profiles_in = load_profiles_data(path, 'core_profiles', dir)
+
+    # Modifying the input config profiles_conditions class
+    core_profiles_conditions = core_profiles_from_IMAS(
+        core_profiles_in, read_psi_from_geo=False
+    )
+    config['geometry']['n_rho'] = 20
+    torax_config = model_config.ToraxConfig.from_dict(config)
+    torax_config.update_fields(core_profiles_conditions)
```
tamaranorman, line 61: Is this needed? We don't necessarily need to override the default core profiles here right?

Can we test the loading path from here separatly

torax/_src/imas_tools/output/tests/core_profiles_test.py
```diff
@@ -0,0 +1,170 @@
...
+    config = self._get_config_dict('test_iterhybrid_rampup_short.py')
+    path = 'core_profiles_ddv4_iterhybrid_rampup_conditions.nc'
+    dir = os.path.join(torax.__path__[0], 'data/third_party/imas_data')
+    core_profiles_in = load_profiles_data(path, 'core_profiles', dir)
+
+    # Modifying the input config profiles_conditions class
+    core_profiles_conditions = core_profiles_from_IMAS(
+        core_profiles_in, read_psi_from_geo=False
+    )
+    config['geometry']['n_rho'] = 20
```
tamaranorman, line 59: This is n_rho anyway

torax/_src/imas_tools/output/tests/core_profiles_test.py
```diff
@@ -0,0 +1,170 @@
...
+from absl.testing import absltest
+from absl.testing import parameterized
+
+try:
+  import imas
+except ImportError:
+  IDSToplevel = Any
+import torax
+from torax._src import state
+from torax._src.imas_tools.input.core_profiles import core_profiles_from_IMAS
```
tamaranorman, line 27: Again import modules not functions

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+  ):  # Map v_loop_lcfs in case it is used as bc for psi equation.
+    v_loop_lcfs = {
+        time_array[ti]: ids.global_quantities.v_loop[ti]
+        for ti in range(len(time_array))
+    }  # TODO: Check the sign for v_loop when it will be used.
+  else:
+    v_loop_lcfs = [0.0]
+
+  return {
+      'profile_conditions.Ip': Ip,
```
tamaranorman, line 128: Do we want to return it like this or more of a nested dict to then use to update the config before it is created as was discussed in the doc

MateoBell:
@tamaranorman This was the initial PR which was split and is not up to date anymore, the input one is #1486. Sorry for the confusion maybe I should either put this one as draft or close it. I will make the changes for the imports on the other PR and will keep the comments regarding the files for the output for followup PR with the output mapping.

Nush395:
@MateoBell, is this functionality in TORAX now?

Can this PR be closed?

MateoBell:
@Nush395 yes it was divided in smaller PR with the last one which we are reviewing.

PR #1477, opened 2025-08-27
Refactor imas coupling

MateoBell:
Move IMAS related functions to live in the same directories. 
Precursor for PR #1443.


torax/_src/imas_tools/imas_utils.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Useful functions for handling of IMAS IDSs."""
+import imas
+from imas import ids_toplevel
+
+
+def load_imas_data(
```
Nush395: If we are making this un-private we usually would have some kind of unit test on it. I feel this is function is so small it's borderline worth just inlining and just calling the two lines:

```
with imas.DBEntry(uri=uri, mode="r") as db:
    ids = db.get(ids_name=ids_name)
```

directly where used, wdyt?

MateoBell: yeah you are right in the end the function is super small, so it is maybe better to inline it and keep a slightly different private version for each IDS type as before.  
Good to know for the unit test I will keep it in mind !

PR #1486, opened 2025-08-29
Initial netCDf core_profiles and coupling functions for input

MateoBell:
Cherry-picked from PR#1443 to split it into smaller pieces. Contains the 2 examples IDSs in netCDF, input transfer function and testing and small documentation (which might be removed and kept for future PR)

To discuss: 

- Usage of Toraxconfig.update_fields or dict manipulation to update the fields ?
- Rudimentary documentation for now, do we keep it or just add documentation with upcoming PR implementing the API to use the coupling ?
- Plasma composition (especially impurity info) kept for follow-up PR
- Removed writing t_initial into the config -> Should it be done or add checks to make sure IMAS profiles loaded are in the simulation time range ?


torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+
+def core_profiles_from_IMAS(
+    ids: IDSToplevel,
+    read_psi_from_geo: bool = True,
+    t_initial: float | None = None,
+) -> dict:
+  """Converts core_profiles IDS to a dict with the input profiles for the config.
+  Args:
+  ids: IDS object. Can be either core_profiles or plasma_profiles. The IDS can
+      contain several time slices.
```
jcitrin: nit: several --> multiple

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+      'profile_conditions.n_e_right_bc': None,
+      'profile_conditions.n_e_nbar_is_fGW': False,
+      'profile_conditions.nbar': None,
+      'profile_conditions.n_e': n_e,
+      'profile_conditions.normalize_n_e_to_nbar': False,
+      'profile_conditions.v_loop_lcfs': v_loop_lcfs,
+  }
+
+
+def load_profiles_data(
```
jcitrin: load_core_profiles_data

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+  in the input config using ToraxConfig.update_fields method.
+  """
+  profiles_1d = ids.profiles_1d
+  time_array = [float(profiles_1d[i].time) for i in range(len(profiles_1d))]
+  if t_initial:
+    time_array = [ti - time_array[0] + t_initial for ti in time_array]
+  rhon_array = [
+      profiles_1d[i].grid.rho_tor_norm for i in range(len(profiles_1d))
+  ]
+  t_initial = float(profiles_1d[0].time)
```
jcitrin: shouldn't this be inside an else block after the `if t_initial:` block?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,155 @@
...
+  # Validator will raise an error if rhon[-1]!= 1.
+  T_e = {
+      time_array[ti]: {
+          rhon_array[ti][rj]: profiles_1d[ti].electrons.temperature[rj] / 1e3
+          for rj in range(len(rhon_array[ti]))
+      }
+      for ti in range(len(time_array))
+  }
+
+  if len(profiles_1d[0].t_i_average > 0):
```
jcitrin: is positivity a sufficient check for existence of t_i_average? Could it also be None, for example? (this depends on the imas python library and conventions, I guess)

torax/_src/imas_tools/input/tests/core_profiles_test.py
```diff
@@ -0,0 +1,156 @@
...
+    path = 'core_profiles_ddv4_iterhybrid_rampup_conditions.nc'
+    dir = os.path.join(torax.__path__[0], 'data/third_party/imas_data')
+    core_profiles_in = load_profiles_data(path, 'core_profiles', dir)
+
+    # Modifying the input config profiles_conditions class
+    core_profiles_conditions = core_profiles_from_IMAS(
+        core_profiles_in, read_psi_from_geo=True
+    )
+    torax_config = model_config.ToraxConfig.from_dict(config)
+    torax_config.update_fields(core_profiles_conditions)
```
jcitrin: this is a case where dict manipulations are more natural. It's not necessary to build a ToraxConfig, just to then immediately update it. Sorry if I may have confused you before on pushing for update_fields no matter what (I may have been confused before).

jcitrin (commented):
I think the main issue here to resolve, is to avoid a situation where IMAS IDS data is loaded into profile_conditions in an "all or nothing" way. The design of `core_profiles_from_IMAS` seems to hint at this. For example, we shouldn't need a `read_psi_from_geo` variable. There should be flexibility in the config, for each variable in profile_conditions, to decide on its input provenance.

To sort this out, we should focus on a brief design doc outlining the specifications of how IMAS data is imported into TORAX. There are many options. For example:

1. Where do we set the datafile(s) to be loaded? How does this look for cases with/without imas-core and an IMASdb?

2. How do we set, per variable, if it's an "IMAS input" or not, and do we set per variable the datafile provenance and t_initial to be associated with that datafile?

3. Where in the internal TORAX stack should the functions that load IMAS data and then "distribute them" to the various Pydantic config objects be implemented and called.

A design doc will answer these questions and then also make the specific PRs and order of PRs clearer. For example, it may make sense to have a precursor PR to this which just deals with setting (in config, maybe in a new "IMAS" config key) the datafiles to load.

jcitrin:
Thanks for this! Did a review.

> Usage of Toraxconfig.update_fields or dict manipulation to update the fields ?

Depends in which context. If, for example as in the tests, we don't _need_ to build a ToraxConfig before modifying some initial template dict, then it's better to do dict manipulation and then build a ToraxConfig. If however, a ToraxConfig has already been built, and we then want to update it, then this is a job for update_fields. I made a comment in the review on the tests.

> Rudimentary documentation for now, do we keep it or just add documentation with upcoming PR implementing the API to use the coupling ?

Can be done in future PR. Basically, documentation just needs to be there by the final PR in a feature chain.

> Plasma composition (especially impurity info) kept for follow-up PR

👍 

> Removed writing t_initial into the config -> Should it be done or add checks to make sure IMAS profiles loaded are in the simulation time range ?

This can be discussed as part of a design doc we should write that clearly lays out the specifications for the IMAS coupling in TORAX. Actually this maybe should have been the first thing we did. Will be in touch separately about this

torax/_src/imas_tools/input/core_profiles.py
MateoBell: IMAS Python requires a strict data type, for this field it should be an array of floats and cannot be None, so I think this check is enough. (modified position of the parenthesis though sorry)

MateoBell:
Ok thanks ! I think I better understand the use cases of update_fields / dict manipulation. Will wait to converge on this design doc before modifying it again to see if we ever need to modify an existing ToraxConfig or not.

MateoBell:
Updated the implementation design following our discussions and the design doc. Could be that the update_dict function is unnecessary, not sure if there is a simpler implementation.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+
+import imas
+from imas import ids_toplevel
+from imas.ids_toplevel import IDSToplevel
+import numpy as np
+import torax
+
+
+def update_dict(old_dict: dict, updates: dict) -> dict:
+  """Recursively modify the fields from the original dict old_dict using the values contained in updates dict.
```
tamaranorman: Keep to 80 chars

And then add whitespace after opening line and before Args and Returns

tamaranorman: Describe the behavior when hitting dicts in more detail (aka if float replacing the whole thing but if strings we are recursively only replacing the new fields)

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+  new_dict = old_dict.copy()
+  for key, value in updates.items():
+    if (
+        isinstance(value, dict)
+        and key in new_dict
+        and isinstance(new_dict[key], dict)
+    ):
+      if all(isinstance(k, float) for k in value.keys()):
+        new_dict[key] = (
+            value  # Needed to replace completely the time slices profiles, instead of keeping the initial ones.
```
tamaranorman: nit: Move comment to above statement to remove whitespace?

```
# Replace completely if keys of the dict are numeric
new_dict[key] = value
```

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+    else:
+      new_dict[key] = value
+  return new_dict
+
+
+def core_profiles_from_IMAS(
+    ids: IDSToplevel,
+    t_initial: float | None = None,
+) -> Mapping[str, Mapping[str, Any]]:
+  """Converts core_profiles IDS to a dict with the input profiles for the config.
```
tamaranorman, line 28: Same comment as above on whitespace

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+  Args:
+  ids: IDS object. Can be either core_profiles or plasma_profiles. The IDS can
+      contain multiple time slices.
+  read_psi_from_geo: Decides either to read psi from the geometry or from the
```
tamaranorman: This PR 1550 should be in soon and will remove the need to have this as the behavior here can be configured by the enum instead of needing to delete the field

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+  ]
+
+  # profile_conditions
+  psi = {
+      t_initial: {
+          rhon_array[0][rj]: profiles_1d[0].grid.psi[rj]
+          for rj in range(len(rhon_array[0]))
+      }
+  }
+  # Will be overwritten if Ip_from_parameters = False, when Ip is given by the equilibrium.
```
tamaranorman: Is this comment needed or if we want it somewhere it should be in documentation

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+      for ti in range(len(time_array))
+  }
+  # It is assumed the temperatures and density profiles are defined until rhon=1.
+  # Validator will raise an error if rhon[-1]!= 1.
+  T_e = {
+      time_array[ti]: {
+          rhon_array[ti][rj]: profiles_1d[ti].electrons.temperature[rj] / 1e3
+          for rj in range(len(rhon_array[ti]))
+      }
+      for ti in range(len(time_array))
```
tamaranorman: If this form of input for the time varying array was used would that be an easier way of setting this up? https://torax.readthedocs.io/en/latest/configuration.html#using-arrays

tamaranorman: And for many of the other params

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+
+def load_core_profiles_data(
+    uri: str,
+    ids_name: str,
+    directory: str | None = None,
+) -> ids_toplevel.IDSToplevel:
+  """Loads a full IDS for a given uri or path_name and a given ids_name which
+  should be either core_profiles or plasma_profiles. It can load either an
```
tamaranorman: Follow https://google.github.io/styleguide/pyguide.html#381-docstrings

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,190 @@
...
+          "n_e_nbar_is_fGW": False,
+          "nbar": None,
+          "n_e": n_e,
+          "normalize_n_e_to_nbar": False,
+          "v_loop_lcfs": v_loop_lcfs,
+      },
+  }
+
+
+def load_core_profiles_data(
```
tamaranorman: Is this method name accurate when it is more load the data described by the ids_name

torax/data/third_party/imas_data/core_profiles_15MA_DT_50_50_flat_top_slice.nc
tamaranorman, line 1: How/why were these generate/where were they used?

torax/_src/imas_tools/input/core_profiles.py
MateoBell: Yes sorry I already removed it from the args but I forgot to remove it in the docstring !

torax/_src/imas_tools/input/core_profiles.py
MateoBell: Yes that's true the function can be generic to any IDS. I think if we agree to put all imas data in data/third_party/imas_data (currently the examples equilibrium IDS are in geo) we can just have one generic loader for any IDS just specifying the name, but I don't know where this function should live.

torax/data/third_party/imas_data/core_profiles_15MA_DT_50_50_flat_top_slice.nc
MateoBell: The provenance of the files were specified in the doctstring of the test where they are used. I added a READme in the folder so that's more visible.

torax/_src/imas_tools/input/core_profiles.py
MateoBell: Indeed I should have pay more attention to the doc, I think it would make the set up easier and the code cleaner, I will try to make the modification tomorrow !

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,194 @@
...
+      profiles_1d[i].grid.rho_tor_norm for i in range(len(profiles_1d))
+  ]
+
+  # profile_conditions
+  psi = {
+      t_initial: {
+          rhon_array[0][rj]: profiles_1d[0].grid.psi[rj]
+          for rj in range(len(rhon_array[0]))
+      }
+  }
```
MateoBell: I just read the initial psi profile from the IDS, as it says "Initial values for poloidal flux." in the doc and it seems to be only read at initialization. Can it be prescribed with time evolution or we only have to provide initial condition like this ?

torax/_src/imas_tools/input/core_profiles.py
tamaranorman: Maybe just in imas_tools/loader.py to keep it simple

torax/_src/imas_tools/input/core_profiles.py
tamaranorman: So you only have to provide an initial condition. You can provide more but only the first will be used

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,194 @@
...
+  Ip = {
+      time_array[ti]: -1 * ids.global_quantities.ip[ti]
+      for ti in range(len(time_array))
```
tamaranorman: Similar to the 2d comment this could be passsed as just time_array, -1 * ids.global_quantities.ip

https://torax.readthedocs.io/en/latest/configuration.html#time-varying-scalars

torax/data/third_party/imas_data/README
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+The IMAS netCDF file core_profiles_ddv4_iterhybrid_rampup_conditions.nc was 
+generated manually filling the core_profiles IDS based on the 
+profile_conditions contained in the iterhybrid_rampup.py example config (after
+unit conversions).
+
+The file core_profiles_15MA_DT_50_50_flat_top_slice.nc comes from a METIS ITER
+baseline scenario time slice taken on the flat top. The IDS grid is made of 21
+points.
```
tamaranorman: nit: Add a newline

torax/_src/imas_tools/input/core_profiles.py
MateoBell: ok as it is now generic to all ids i removed the _load_geo_data from equilibrium and moved the equilibrium netCDF IDSs to the imas_data folder !

docs/configuration.rst
```diff
@@ -2037,6 +2036,27 @@ There are three main methods to load IMAS equilibrium:
...
   Using this method the IDS can be provided externally or pre-loaded.
 
+Loading Profiles
+^^^^^^^^^^^^^^^^
+
+Profiles in TORAX can be read from any IMAS core_profiles or plasma_profiles IDS
+saved in Data Dictionary version 4.0.0 or newer.
+If the IDS is stored in an IMAS db or in a netCDF file it can be loaded using the
+loader function ``load_core_profiles_data`` from |core_profiles_input_imas|.
+It can then be loaded programatically in the CONFIG by constructing a nested
```
Nush395: nit:  shall we use ``CONFIG`` here?

docs/configuration.rst
```diff
@@ -2037,6 +2036,27 @@ There are three main methods to load IMAS equilibrium:
...
+dictionary with the ``core_profiles_from_imas`` function. The function returns 
+a dictionary whose structure fits the schema of profile_conditions and can be 
+programatically loaded into a CONFIG with standard dictionary manipulation
```
Nush395: nit:  shall we use ``CONFIG`` here?

torax/_src/imas_tools/input/loader.py
```diff
@@ -0,0 +1,46 @@
...
+
+  It can load either an IMAS netCDF file with filename as uri and given
+  directory or from an IMASdb by giving the full uri of the IDS and the
+  directory arg will be ignored. Note that loading from an IMASdb requires
+  IMAS-core. The loaded IDS can then be used as input to
+  core_profiles_from_IMAS().
+  """
+  # Differentiate between netCDF and IMASdb uris. For IMASdb files the full
+  # filepath is already provided in the uri.
+  if uri[-3:] == ".nc":
```
Nush395: nit: perhaps use `uri.endswith(".nc")` here to read a bit better

torax/_src/imas_tools/input/tests/equilibrium_test.py
```diff
@@ -77,8 +77,10 @@ def test_IMAS_input_with_filepath(self):
 
   def test_IMAS_input_with_uri(self):
     filename = 'ITERhybrid_COCOS17_IDS_ddv4.nc'
-    geometry_directory = geometry_loader.get_geometry_dir()
-    full_path = os.path.join(geometry_directory, filename)
+    imas_directory = os.path.join(
+        torax.__path__[0], 'data/third_party/imas_data'
```
Nush395, line 81: perhaps here instead we can use 

```
from torax._src.config import config_loader

config_loader.torax_path()
```

I believe that fits better with our internal systems as well.

torax/_src/imas_tools/input/tests/equilibrium_test.py
```diff
@@ -91,7 +93,7 @@ def test_IMAS_input_with_uri(self):
 
   def test_IMAS_input_with_equilibrium_object(self):
     filename = 'ITERhybrid_COCOS17_IDS_ddv4.nc'
-    equilibrium_in = imas_geometry._load_geo_data(filename)
+    equilibrium_in = loader.load_imas_data(filename, 'equilibrium')
```
Nush395: we could introduce some named constants in `loader` for the different `ids_names`, wdyt?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,160 @@
...
+"""Useful functions to load IMAS core_profiles or plasma_profiles IDSs and
+converts them into TORAX objects.
+"""
+from typing import Any, Mapping
+
+from imas import ids_toplevel
+import numpy as np
+
+
+def update_dict(old_dict: dict, updates: dict) -> dict:
```
Nush395: is this intended to be used within TORAX or is it more a helper for tests?

torax/_src/imas_tools/input/tests/core_profiles_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+    # Modifying the input config profiles_conditions class
+    core_profiles_conditions = core_profiles.core_profiles_from_IMAS(
+        core_profiles_in,
+    )
+    config = core_profiles.update_dict(config, core_profiles_conditions)
+    torax_config = model_config.ToraxConfig.from_dict(config)
+
+    # Run Sim
+    _, results = torax.run_simulation(torax_config)
```
Nush395: what is the purpose of running the whole simulation here?

docs/configuration.rst
MateoBell: What do you suggest instead ? Is it possible to like load the profile_conditions into TORAX through another object ?

torax/_src/imas_tools/input/tests/equilibrium_test.py
MateoBell: Yes good idea ! Can you tell me if I implemented it in a proper way ?

torax/_src/imas_tools/input/core_profiles.py
MateoBell: I think it can be seen as a helper for tests, useful only to update multiples dict from the config at once (for when plasma_composition will be mapped as well by the core_profiles converter for example).
For other use cases, loading all imas outputs into a config with e.g. `config['profile_conditions']={**imas_data.profile_conditions,}` should be practical enough. 

Nevertheless thinking a bit about it it might not even be that necessary for tests so I could supress it if you prefer. 
Wdyt on this ?

torax/_src/imas_tools/input/tests/core_profiles_test.py
MateoBell: Just an additional test, which might be pretty similar to a sim_test in the end. Do you want me to remove it ?

docs/configuration.rst
Nush395: sorry I just meant in the doc to format it as CONFIG in between the backticks ```` to format as "code"

torax/_src/imas_tools/input/core_profiles.py
Nush395: I think if it's useful for tests it could live in the test file and be a helper there but if you think it's not necessary for tests either then I'm in favour of removing it too!

torax/_src/imas_tools/input/tests/core_profiles_test.py
Nush395: if I understand the behaviour of the `core_profiles_from_IMAS` is tested in the test below `test_init_profiles_from_IMAS` and then after that the results of the simulation should be the same so they would be covered by existing tests. Is that correct?

If so I think it could be removed

torax/_src/imas_tools/input/tests/equilibrium_test.py
Nush395: that looks good!

docs/configuration.rst
MateoBell: ah okay sorry didnt pay enough attention ahah !

torax/_src/imas_tools/input/core_profiles.py
MateoBell: I removed it, with regular dict manipulation it seems clean enough I think !

torax/_src/imas_tools/input/tests/core_profiles_test.py
MateoBell: Yes I agree, just added the test to make sure the initial profiles didn't make the simulation crash but that should be covered by existing tests and the sim_test that will be added in followup PR

docs/configuration.rst
```diff
@@ -2037,6 +2036,27 @@ There are three main methods to load IMAS equilibrium:
 * Providing the equilibium IDS on the fly (equilibrium_object).
   Using this method the IDS can be provided externally or pre-loaded.
 
+Loading Profiles
+^^^^^^^^^^^^^^^^
+
+Profiles in TORAX can be read from any IMAS core_profiles or plasma_profiles IDS
+saved in Data Dictionary version 4.0.0 or newer.
+If the IDS is stored in an IMAS db or in a netCDF file it can be loaded using the
+loader function ``load_core_profiles_data`` from |core_profiles_input_imas|.
```
Nush395: is this the function now called `core_profiles_from_IMAS` ?

torax/_src/imas_tools/input/loader.py
```diff
@@ -0,0 +1,53 @@
...
+import imas
+from imas import ids_toplevel
+from torax._src.config import config_loader
+
+# Names of the IDSs that can be loaded and used in TORAX.
+CORE_PROFILES: Final[str] = "core_profiles"
+EQUILIBRIUM: Final[str] = "equilibrium"
+
+
+def load_imas_data(
```
Nush395, line 31: Can we have a unit test for this function? Checking that we can load via the netcdf pathway.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+  if t_initial:
+    time_array = [ti - time_array[0] + t_initial for ti in time_array]
+  else:
+    t_initial = float(profiles_1d[0].time)
+  rhon_array = [
+      profiles_1d[i].grid.rho_tor_norm for i in range(len(profiles_1d))
+  ]
+
+  # profile_conditions
+  psi = (np.array(rhon_array[0]), np.array(profiles_1d[0].grid.psi))
```
Nush395, line 52: why do we save the `np.array(rhon_array[0])` here?

Nush395: Ah ok, this is to match one of the shortcuts for initialising a `TimeVaryingArray` is that correct?

Why is the `psi` assumed to be constant?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+        initial time will be the time of the first time slice of the ids. Else
+        all time slices will be shifted such that the first time slice has
+        time = t_initial.
+
+  Returns:
+     Dict containing the updated fields read from the IDS that need to be replaced
+     in the input config using ToraxConfig.update_fields method.
+  """
+  profiles_1d = ids.profiles_1d
+  time_array = [float(profiles_1d[i].time) for i in range(len(profiles_1d))]
```
Nush395: is this `float` cast necessary?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+    t_initial = float(profiles_1d[0].time)
+  rhon_array = [
+      profiles_1d[i].grid.rho_tor_norm for i in range(len(profiles_1d))
+  ]
+
+  # profile_conditions
+  psi = (np.array(rhon_array[0]), np.array(profiles_1d[0].grid.psi))
+  Ip = (
+      time_array,
+      -1 * ids.global_quantities.ip,
```
Nush395, line 57: can you add a comment here why we multiply by `-1`?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+  )
+  # It is assumed the temperatures and density profiles are defined until rhon=1.
+  # Validator will raise an error if rhon[-1]!= 1.
+  T_e = (
+      time_array,
+      rhon_array,
+      [
+          profiles_1d[ti].electrons.temperature / 1e3
```
Nush395: can you add a comment about the unit conversion here

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+          for ti in range(len(time_array))
+      ],
+  )
+
+  if len(profiles_1d[0].t_i_average) > 0:
+    T_i = (
+        time_array,
+        rhon_array,
+        [profiles_1d[ti].t_i_average / 1e3 for ti in range(len(time_array))],
```
Nush395: can you add a comment about the unit conversion here

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+  else:
+    t_i_average = [
+        np.mean(
+            [
+                profiles_1d[ti].ion[iion].temperature
+                for iion in range(len(profiles_1d[ti].ion))
```
Nush395: nit: type `iion` -> `ion` ?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -0,0 +1,123 @@
...
+  )
+
+  # Map v_loop_lcfs in case it is used as bc for psi equation.
+  if len(ids.global_quantities.v_loop) > 0:
+    v_loop_lcfs = (
+        time_array,
+        ids.global_quantities.v_loop,
+    )  # TODO: Check the sign for v_loop when it will be used.
+  else:
+    v_loop_lcfs = [0.0]
```
Nush395: can this just be `0.0` ?

Nush395 (approved):

PR #1488, opened 2025-09-01
Add <B^2> and <1/B^2> geometry terms

theo-brown:
- [x] Circular geometry
- [x] CHEASE
- [x] IMAS
- [ ] FBT
- [ ] EQDSK


theo-brown:
@jcitrin / @felicif I only just requested access to the MEQ repo, so I can't check this - asking here instead: is there a definition of epsilon (inverse aspect ratio) in the FBT outputs, or should I compute it manually?

jcitrin:
@theo-brown: MEQ/FBT does have an LY['epsilon'] output. However in TORAX, epsilon (and epsilon_face) are [properties](https://github.com/google-deepmind/torax/blob/164eaefb7ecfb0228ea17eeee368aab9ca021ceb/torax/_src/geometry/geometry.py#L274) of the Geometry dataclass. The definition used in the property definition is the same definition as in MEQ. So can just use that, no need for extra calculations

theo-brown:
Sense-check (gm4 left, gm5 right)
gm4_gm5

 @jcitrin I can add the comparison with the analytical formula if required, but I think the current test isn't too bad - it just checks that CHEASE, IMAS and EQDSK are consistent. As we are only manually computing it for EQDSK, I feel like this is enough of a check?

torax/_src/geometry/tests/standard_geometry_test.py
```diff
@@ -86,6 +88,25 @@ def test_build_standard_geometry_from_IMAS(self):
...
+    ).build_geometry()
+
+    chease_geo = geometry_pydantic_model.CheaseConfig(
+        geometry_file='ITER_hybrid_citrin_equil_cheasedata.mat2cols'
+    ).build_geometry()
+
+    np.testing.assert_allclose(eqdsk_geo.gm4, chease_geo.gm4, atol=1e-3)
+    np.testing.assert_allclose(imas_geo.gm4, chease_geo.gm4, atol=1e-3)
+    np.testing.assert_allclose(eqdsk_geo.gm5, chease_geo.gm5, atol=0.5)
+    np.testing.assert_allclose(imas_geo.gm5, chease_geo.gm5, atol=0.5)
```
jcitrin: nit: better to use rtol (unless values are really close to 0) since it's more intuitive to glean at a glance how close they are

jcitrin (approved):

jcitrin:
> @jcitrin I can add the comparison with the analytical formula if required, but I think the current test isn't too bad - it just checks that CHEASE, IMAS and EQDSK are consistent. As we are only manually computing it for EQDSK, I feel like this is enough of a check?

Yes this looks good, thanks! We can merge after the nit and the conflicts are sorted

torax/_src/geometry/tests/standard_geometry_test.py
theo-brown: Good point, thanks. I was being a bit lazy! Set it to the lowest 1 s.f. rtol that passes.

theo-brown:
Thanks! I'll add the changes to the Angioni-Sauter model now.

PR #1489, opened 2025-09-01
Switch psi sources to being <j.B>

theo-brown:
- [x] Change total_psi_sources
- [ ] Change individual psi sources:
  - [x] ECRH
  - [ ] Generic current source
- [ ] Regenerate tests


theo-brown:
@jcitrin am I right that the noninductive psi sources are ECCD, bootstrap, and generic current source? Am I missing any?

ECCD is field-aligned following #1363. Bootstrap is field aligned by definition?

Currently the generic current source is set to provide a fixed toroidal current in A (or fraction of the total current). Would you like this to remain as specifying a toroidal current, or is it ok to silently change it to being the specification of a field-aligned current? Can add to the generic current source either documentation saying it's a <j.B> current or add the conversion from j_tor to <j.B>.

torax/_src/sources/source_profiles.py
```diff
@@ -121,11 +121,14 @@ def merge(
...
+    # All psi sources are assumed to be parallel to the magnetic field, ie
+    # self.psi.values() is <j.B> / B0
+    total_j_dot_B_over_B0 = self.bootstrap_current.j_bootstrap
+    total_j_dot_B_over_B0 += sum(self.psi.values())
+    total_j_dot_B *= geo.B_0
     mu0 = constants.CONSTANTS.mu0
-    prefactor = 8 * geo.vpr * jnp.pi**2 * geo.B_0 * mu0 * geo.Phi_b / geo.F**2
-    return -total * prefactor
+    prefactor = 8 * geo.vpr * jnp.pi**2 * mu0 * geo.Phi_b / geo.F**2
+    return -total_j_dot_B * prefactor
```
theo-brown, line 131: Minor rearranging/renaming for readability and to match equation in paper.

torax/_src/sources/source_profiles.py
```diff
@@ -121,11 +121,14 @@ def merge(
...
+    # All psi sources are assumed to be parallel to the magnetic field, ie
+    # self.psi.values() is <j.B> / B0
+    total_j_dot_B_over_B0 = self.bootstrap_current.j_bootstrap
+    total_j_dot_B_over_B0 += sum(self.psi.values())
+    total_j_dot_B = total_j_dot_B_over_B0 * geo.B_0
     mu0 = constants.CONSTANTS.mu0
-    prefactor = 8 * geo.vpr * jnp.pi**2 * geo.B_0 * mu0 * geo.Phi_b / geo.F**2
-    return -total * prefactor
+    prefactor = 8 * geo.vpr * jnp.pi**2 * mu0 * geo.Phi_b / geo.F**2
+    return -total_j_dot_B * prefactor
```
theo-brown, line 131: This is mostly cosmetic stuff, just wanted to make it clear to anyone reading what was going on and for the equation to match the one in the paper.

jcitrin:
> @jcitrin am I right that the noninductive psi sources are ECCD, bootstrap, and generic current source? Am I missing any?

For now yes. In principle there could be NBI or Lower Hybrid (LH) current sources, but we don't have models for them yet.

> Bootstrap is field aligned by definition?

Yes, the output of the bootstrap model is <j \cdot B>/B_0 . So its usage in the psi equation is correct. However, what is incorrect is our _interpretation_ elsewhere that it is flux-surface-averaged toroidal current. For example in the output I_p, I_bootstrap, I_ecrh, everything is assumed as toroidal current. In any internal calculations calculating j_ohmic etc., for the initialization stuff, then we assume that all currents are always toroidal.

What we need are conversion formulas from <j \cdot B>/B_0 to j_torax = <j_phi / R> / <1/R> , and vice versa. I'll make a separate issue for this (even though it's intimately linked to this issue).

> Currently the generic current source is set to provide a fixed toroidal current in A (or fraction of the total current). Would you like this to remain as specifying a toroidal current, or is it ok to silently change it to being the specification of a field-aligned current? Can add to the generic current source either documentation saying it's a <j.B> current or add the conversion from j_tor to <j.B>.

I think the generic current source input description should remain "toroidal current" which is more user intuitive. Then internally we have the conversions to what's need to the psi equation, as discussed.

Thanks for being on top of this one!

theo-brown:
Cool, I'll implement and test the conversion in #1508 and then hunt down all the places it needs to be converted in here.

torax/_src/physics/psi_calculations.py
```diff
@@ -338,3 +341,47 @@ def calculate_psidot_from_psi_sources(
   psidot = (jnp.dot(c_mat, psi.value) + c) / toc_psi
 
   return psidot
+
+
+def j_tor_to_j_parallel(
+    j_tor: array_typing.FloatVector, geo: geometry.Geometry
+) -> array_typing.FloatVector:
```
theo-brown: I haven't checked face vs cell-ness. Will this always be called with j on the cell grid? Or do I need to create either a separate function or internal branching for face calculation.

torax/_src/physics/tests/psi_calculations_test.py
```diff
@@ -190,6 +190,28 @@ def test_calc_Wpol(self):
     # our circular geometry, but approximates it at low inverse aspect ratio.
     np.testing.assert_allclose(calculated_Wpol, expected_Wpol, rtol=1e-3)
 
+  # pylint: enable=invalid-name
+
+  def test_calc_j_tor_from_j_parallel(self):
+    geo = None
+    j_tor_truth = None
+    j_parallel_truth = None
```
theo-brown: What should I have as the sources of truth from in this test? Concerned I might end up testing it circularly!

theo-brown:
Currently, this is set up to have `core_sources.psi` as containing current densities in `<j.B>/B0` form. This makes sense to me as it matches what the sources in the psi equation are expected to be.

My current plan is to add toroidal current densities to `PostProcessedOutputs` and then explicitly state in the output file which is which. Does `j_toroidal_ecrh` vs `j_parallel_ecrh` as a naming convention for the outputs sound ok? 
The outputs would look like this:

image

jcitrin:
The convention _parallel_ vs _toroidal_ is good.

However, we have committed not to make breaking changes to output API until V2. This means that new additions can be made, but name changes cannot. So I recommend having something like j_foo_parallel and j_foo, where j_foo is toroidal, and this is mentioned explicitly in the docs. A TODO for name updates conversion to V2 can be made.

torax/_src/sources/generic_current_source.py
```diff
@@ -57,8 +58,9 @@ def calculate_generic_current(
     unused_state: state.CoreProfiles,
     unused_calculated_source_profiles: source_profiles.SourceProfiles | None,
     unused_conductivity: conductivity_base.Conductivity | None,
-) -> tuple[array_typing.FloatVectorCell, ...]:
-  """Calculates the external current density profiles on the cell grid."""
+) -> tuple[chex.Array, ...]:
```
jcitrin: no need to change the output typehint

torax/_src/physics/tests/psi_calculations_test.py
jcitrin: Is it possible to run with one of our EQDSK or IMAS equilibrium inputs in both JETTO and TORAX, and verify a conversion "by eye", and then after doing that just put the j_parallel_truth and j_toroidal_truth (for a small grid) by hand in this file, and load that same geo? Then at least it's a regression test.

torax/_src/physics/psi_calculations.py
jcitrin: For now this is only needed for the source calculations where it's just on the cell grid. So OK to keep like this for now. If we'll want on the face grid too we can extend later

torax/_src/output_tools/post_processing.py
```diff
@@ -135,9 +135,18 @@ class PostProcessedOutputs:
...
+    j_parallel_ohmic: Parallel ohmic current density [Am^-2]
+    j_toroidal_total: Total toroidal current density [Am^-2]
+    j_toroidal_bootstrap: Toroidal bootstrap current density [Am^-2]
+    j_toroidal_ohmic: Toroidal ohmic current density [Am^-2]
+    j_toroidal_external: Toroidal current density from external psi sources
+      (i.e., excluding bootstrap) [A m^-2]
+    j_toroidal_generic_current: Toroidal current density from generic current source
+      [Am^-2]
+    j_toroidal_ecrh: Toroidal current density from electron cyclotron heating
+      and current source [Am^-2]
```
jcitrin, line 161: The only constraint we have is not to change existing names. So it's good to add these extra terms but perhaps just add either j_toroidal_foo (and have all the j_foo be parallel) or vice-versa, whichever makes more sense. Can add a TODO to update the names to both _toroidal and _parallel in V2

torax/_src/sources/generic_current_source.py
theo-brown: Thanks, it was an accident!

torax/_src/physics/tests/psi_calculations_test.py
theo-brown: Yep, can do! I'll do it in the next couple of days.

theo-brown:
> Is it possible to run with one of our EQDSK or IMAS equilibrium inputs in both JETTO and TORAX, and verify a conversion "by eye", and then after doing that just put the j_parallel_truth and j_toroidal_truth (for a small grid) by hand in this file, and load that same geo? Then at least it's a regression test.

Turns out JETTO also has inconsistencies in its IMAS outputs with <j.B> vs j_toroidal, so we're fixing that upstream. Once that's sorted, we'll be able to do the comparison easily!

theo-brown:
Comparison JETTO/TORAX, OpenSTEP case. 

[j_tor -> j_parallel conversion](https://github.com/google-deepmind/torax/blob/95238ee28df02dd201fd91d3e9ae2efcf5f18752/torax/_src/physics/psi_calculations.py#L346)
image

[j_parallel -> j_tor conversion](https://github.com/google-deepmind/torax/blob/95238ee28df02dd201fd91d3e9ae2efcf5f18752/torax/_src/physics/psi_calculations.py#L380)
image

theo-brown:
`iterhybrid_predictor_corrector`, comparison of treating current sources and j_bootstrap as parallel (solid) with the version on main (dashed)
Figure_1


As expected, the difference is more pronounced in `step_flattop_bgb`. Most important difference is the impact on bootstrap current fraction. q profile is unaffected.


Figure_1

theo-brown:
I've added a unit test that's comparing against the ground truth in the STEP IMAS case that's now included in the repo.

theo-brown:
One last bug to squash regarding on-axis values, then ready for review
image

theo-brown:
@jcitrin actually would appreciate your scientific computing expertise here. I've tried a handful of tricks (e.g. replacing `1/rho dX/drho -> 2 dX/drho^2`, linearly interpolating the last few values, extending with a cell at rho=0, etc) but they all break when the grid size is adjusted.

What's the most TORAX-native way of handling the axis boundary condition?

torax/_src/physics/tests/psi_calculations_test.py
```diff
@@ -190,6 +192,58 @@ def test_calc_Wpol(self):
     # our circular geometry, but approximates it at low inverse aspect ratio.
     np.testing.assert_allclose(calculated_Wpol, expected_Wpol, rtol=1e-3)
 
+  # pylint: enable=invalid-name
+
+  def test_calc_j_tor_from_j_parallel(self, step_geo_and_j_profiles_from_IMAS):
```
theo-brown: This could potentially be moved to test_utils/torax_refs. However, it's a bit of a different setup to the code that's already there, as we're interested in j_toroidal vs j_parallel rather than the psi, psidot, q, s profiles.

torax/_src/output_tools/post_processing.py
```diff
@@ -146,9 +146,24 @@ class PostProcessedOutputs:
...
+      (i.e., excluding bootstrap) [A m^-2]
+    j_ohmic: Toroidal ohmic current density [Am^-2]
+    j_parallel_ohmic: Parallel ohmic current density [Am^-2]
+    j_generic_current: Toroidal current density from generic current source
+      [Am^-2]
+    j_parallel_generic_current: Parallel current density from generic current
+      source [Am^-2]
+    j_ecrh: Toroidal current density from electron cyclotron heating
+      and current source [Am^-2]
+    j_paralell_ecrh: Toroidal current density from electron cyclotron heating
```
jcitrin: nit: typo

torax/_src/physics/psi_calculations.py
```diff
@@ -26,15 +26,15 @@
     - calculate_psi_grad_constraint_from_Ip: Calculates the gradient
       constraint on the poloidal flux (psi) from Ip.
     - _calc_bpol2: Calculates square of poloidal field (Bp).
+      constraint on the poloidal flux (psi) from Ip.
+    - j_tor_to_j_parallel: Calculates <j.B>/B0 from j_tor = dI/dS.
+    - j_parallel_to_j_tor: Calculates j_tor = dI/dS from <j.B>/B0.
```
jcitrin: nit: tor --> toroidal

torax/_src/physics/tests/psi_calculations_test.py
jcitrin: I think it's fine to leave as is

torax/_src/physics/tests/psi_calculations_test.py
```diff
@@ -190,6 +192,58 @@ def test_calc_Wpol(self):
...
+    np.testing.assert_allclose(
+        psi_calculations.j_parallel_to_j_toroidal(j_parallel_from_j_tor, geo),
+        j_tor_truth,
+        rtol=1e-6,
+    )
+
+
+@pytest.fixture
+def step_geo_and_j_profiles_from_IMAS():
+  # TODO: move to torax_refs
```
jcitrin: can remove the TODO. Maybe replace with a reminder that the .nc file used is in the default `data/imas_data` directory

torax/_src/physics/psi_calculations.py
```diff
@@ -338,3 +338,64 @@ def calculate_psidot_from_psi_sources(
...
+        - \frac{\partial F}{\partial \rho} I
+      \right)
+
+  where :math:`\frac{\partial I}{\partial \rho} =
+    j_{\mathrm{tor}} \frac{\partial S}{\partial \rho}`.
+
+  See Eq 45 in https://users.euro-fusion.org/pages/data-cmg/wiki/files/JETTO_current.pdf
+  """
+  dI_drhon = j_toroidal * geo.spr
+  I = math_utils.cumulative_trapezoid(dI_drhon * geo.drho_norm, initial=0.0)
```
jcitrin: For cell grid area integrations we should use `math_utils.area_integration` . Then these two lines will just be

`I = math_utils.area_integration(j_torodial, geo)`

This treats the cell variables as averaged cell variables as they should be, and the integration result at idx=0 will not be 0.0.

the cell_to_face utility then maintains the surface integral when setting `preserved_quantity=math_utils.IntegralPreservationQuantity.SURFACE` , which is what you have done correctly in this PR.

Maybe fixing this will help with the on-axis issue?

torax/_src/physics/psi_calculations.py
```diff
@@ -338,3 +338,64 @@ def calculate_psidot_from_psi_sources(
...
+
+  Operates only on the cell grid.
+
+  The relationship is
+
+  .. math::
+    j_\mathrm{tor} = \frac{\partial}{\partial S} \left(\frac{2 \pi B_0^2}{F} \int_{0}^{\rho'}
+    \frac{\langle j.B \rangle}{B_0 F \langle 1/R^2 \rangle} \rho' d\rho'\right)
+
+  See Eq 57 https://users.euro-fusion.org/pages/data-cmg/wiki/files/JETTO_current.pdf
```
jcitrin, line 462: could these notes be made open somewhere?

torax/_src/physics/psi_calculations.py
```diff
@@ -338,3 +338,64 @@ def calculate_psidot_from_psi_sources(
...
+  """
+  I = (
+      2
+      * jnp.pi
+      * geo.B_0**2
+      * geo.F
+      * math_utils.cumulative_trapezoid(
```
jcitrin: Same comment as above on the use of math_utils.area_integration

torax/_src/physics/psi_calculations.py
```diff
@@ -338,3 +338,64 @@ def calculate_psidot_from_psi_sources(
...
+          j_parallel * geo.rho / (geo.F**3 * geo.g3) * geo.drho, initial=0.0
```
jcitrin: There seems to be a bit of an inconsistency with F compared to the docstring?

In the docstring there's 1/F outside the integral, and 1/F inside the integral. In the actual implementation there's F outside the integral, and 1/F^3 inside the integral

torax/_src/physics/psi_calculations.py
theo-brown: Not sure what's going on here, but switching to area_integration breaks things:

image

torax/_src/physics/psi_calculations.py
theo-brown: I have been trying, I will push again

torax/_src/physics/psi_calculations.py
theo-brown: This would be `math_utils.cell_integration`, if I understand correctly? And then multiply by the Jacobian `drho/drho_norm` to get the integral wrt rho. Unless you're suggesting changing the `rho*drho`?

theo-brown: This would be math_utils.cell_integration, if I understand correctly? And then multiply by the Jacobian drho/drho_norm to get the integral wrt rho. Unless you're suggesting changing the rho*drho?

torax/_src/physics/psi_calculations.py
theo-brown: I have been trying, I will do some more poking and see whether I can move it along.

torax/_src/physics/psi_calculations.py
theo-brown: Not sure what's going on here, but switching to area_integration breaks things:
image

jcitrin: Sorry I made an error. `math_utils.area_integration` is not a cumulative sum. I just made cumulative versions of `cell_integration`, `area_integration`, and `volume_integration` in https://github.com/google-deepmind/torax/pull/1745 , they are `cumulative_cell_integration`, `cumulative_area_integration`, `cumulative_volume_integration`. You can try merging that branch (or waiting until it's soon merged and sync to main) and trying with `cumulative_area_integration`.

The input arguments are just the profile you want to integrate, and geo. The area and volume variants internally then multiply by the area or volume metric.

These integrals are Riemann sums as opposed to using the trapezoidal rule. This is more appropriate for integrals on the cell grid where each value is interpreted as a cell average. With these new functions, the first element will not be zero since it's just the rectangle geo.drho_norm*value[0].

I checked offline and for the `iterhybrid_rampup` case at the first time point, these are the differences between
`I_trapezoid = math_utils.cumulative_trapezoid( j_toroidal * geo.spr * geo.drho_norm, initial=0.0)`
```
Array([0.00000000e+00, 2.79029752e+00, 3.84262802e+01, 3.25437047e+02,
       1.93383752e+03, 8.45333219e+03, 2.79437684e+04, 7.14336130e+04,
       1.44440876e+05, 2.37124244e+05, 3.26400024e+05, 3.91767598e+05,
       4.28169880e+05, 4.43581147e+05, 4.48535265e+05, 4.49742607e+05,
       4.49965318e+05, 4.49996366e+05, 4.49999632e+05, 4.49999891e+05,
       4.49999907e+05, 4.49999908e+05, 4.49999908e+05, 4.49999908e+05,
       4.49999908e+05], dtype=float64)
```

and

`I_riemann = math_utils.cumulative_area_integration(j_toroidal, geo)`
```
Array([1.84830790e-01, 5.58059504e+00, 7.14567961e+01, 5.79602129e+02,
       3.28825774e+03, 1.36185915e+04, 4.22691301e+04, 1.00598281e+05,
       1.88283657e+05, 2.85965016e+05, 3.66835216e+05, 4.16700164e+05,
       4.39639780e+05, 4.47522699e+05, 4.49548016e+05, 4.49937384e+05,
       4.49993436e+05, 4.49999481e+05, 4.49999969e+05, 4.49999999e+05,
       4.50000000e+05, 4.50000000e+05, 4.50000000e+05, 4.50000000e+05,
       4.50000000e+05], dtype=float64)
```

torax/_src/physics/psi_calculations.py
jcitrin: see above comment on the new `math_utils.cumulative_area_integration`

torax/_src/physics/psi_calculations.py
theo-brown: Nice! Will retry with those now

theo-brown:
I've rebased onto #1745. On-axis behaviour is still a bit unexpected. 
image

torax/_src/physics/psi_calculations.py
jcitrin: The new functions are now merged to main

jcitrin:
OK will look deeper at the on-axis values. Could you also resync again to main to resolve the conflicts following the latest update?

jcitrin:
Regarding the plot, are both `j_toroidal` and `j_parallel` shown after the round trip? As in, does the "original" `j_toroidal` also have the on-axis explosion or is this isolated to the new functions?

jcitrin:
Also how can I best reproduce this with data currently in the repo?

theo-brown:
Both are computed directly from the corresponding IMAS entry. So `j_parallel_torax = j_toroidal_to_j_parallel(j_toroidal_imas, geo_from_imas)` and `j_toroidal_torax = j_parallel_to_j_toroidal(j_parallel_imas, geo_from_imas)`

Repro
```python
import matplotlib.pyplot as plt 
from torax._src.geometry import pydantic_model as geometry_pydantic_model
from torax._src.physics import psi_calculations
import jax.numpy as jnp
from torax._src import math_utils
from torax._src.imas_tools.input import loader as imas_loader

# Build geo
config = geometry_pydantic_model.IMASConfig(
    imas_filepath='STEP_SPP_001_ECHD_ftop.nc',
    n_rho=250,
)
geo = config.build_geometry()

# Get ground truth profiles
core_profiles_ids = imas_loader.load_imas_data('STEP_SPP_001_ECHD_ftop.nc', 'core_profiles')
rho_norm_imas =  jnp.asarray(core_profiles_ids.profiles_1d[0].grid.rho_tor_norm)
j_toroidal_imas = jnp.asarray(core_profiles_ids.profiles_1d[0].j_phi)
j_parallel_imas = jnp.asarray(core_profiles_ids.profiles_1d[0].j_total)
j_toroidal_imas = jnp.interp(geo.rho_norm, rho_norm_imas, j_toroidal_imas)
j_parallel_imas = jnp.interp(geo.rho_norm, rho_norm_imas, j_parallel_imas)

# Plot
plt.plot(geo.rho_norm, j_parallel_imas, color="C0", label="j_parallel (JETTO)")
plt.plot(geo.rho_norm, psi_calculations.j_toroidal_to_j_parallel(j_toroidal_imas, geo), color="C0", linestyle="--", linewidth=3, label="j_parallel (TORAX)")
plt.plot(geo.rho_norm, j_toroidal_imas, color="C1", label="j_toroidal (JETTO)")
plt.plot(geo.rho_norm, psi_calculations.j_parallel_to_j_toroidal(j_parallel_imas, geo), color="C1", linestyle="--", linewidth=3, label="j_toroidal (TORAX)")
plt.legend()
```

jcitrin:
One issue that came up internally is the new folder "third_party" naming breaks some directory naming conventions. Therefore will just move everything there back to `data/imas`  (not `data/imas_data` which is a redundant name).

No action needed, just a fyi

torax/_src/physics/psi_calculations.py
theo-brown: It's stuck in internal review atm, I have asked for it to be bumped up the priority list.

torax/_src/physics/psi_calculations.py
```diff
@@ -338,3 +337,64 @@ def calculate_psidot_from_psi_sources(
...
+    j_parallel: array_typing.FloatVector, geo: geometry.Geometry
+) -> array_typing.FloatVector:
+  r"""Calculates j_toroidal = dI/dS from <j.B>/B0.
+
+  Operates only on the cell grid.
+
+  The relationship is
+
+  .. math::
+    j_\mathrm{tor} = \frac{\partial}{\partial S} \left(\frac{2 \pi B_0^2}{F} \int_{0}^{\rho}
```
jcitrin: typo: it's not a fraction, just 2*pi*B^2*F

jcitrin:
After taking a look, it seems that an important aspect is to calculate _face  grid_ terms for differentiation onto the cell grid,  analogous to how we treat the PDE diffusion-convection terms in the finite volume method. This improves accuracy for the bulk. Then, for the on-axis treatment, just constant extrapolation below some small cutoff looks sufficient. See example code below

```
_CUTOFF_RHO = 0.05

def _extrapolate_to_axis(
    profile: array_typing.FloatVector,
    geo: geometry.Geometry,
    cutoff_rho: float = _CUTOFF_RHO,
) -> array_typing.FloatVector:
  """Extrapolates the core value to avoid on-axis numerical artifacts."""

  # Interpolate to find the value at exactly rho_norm = cutoff_rho
  val_at_cutoff = jnp.interp(cutoff_rho, geo.rho_norm, profile)

  # Clamp all values below cutoff to this value
  return jnp.where(geo.rho_norm < cutoff_rho, val_at_cutoff, profile)


def j_toroidal_to_j_parallel(
    j_toroidal: array_typing.FloatVectorCell, geo: geometry.Geometry
) -> array_typing.FloatVectorCell:
  r"""Calculates <j.B>/B0 from j_tor = dI/dS.

  Operates only on the cell grid. 
  
  The relationship is .. math::

    \frac{\langle j.B \rangle}{B_0} =
      \frac{F \langle 1/R^2 \rangle}{2 \pi \rho B_0^2}
      \left(
        F \frac{\partial I}{\partial \rho}
        - \frac{\partial F}{\partial \rho} I
      \right)
  
  where :math:`\frac{\partial I}{\partial \rho} =
    j_{\mathrm{tor}} \frac{\partial S}{\partial \rho}`.
  
  See Eq 45 in
  https://users.euro-fusion.org/pages/data-cmg/wiki/files/JETTO_current.pdf

  Args:
    j_toroidal: Toroidal current density profile on the cell grid.
    geo: Geometry object.

  Returns:
    Parallel current density, <j.B>/B0, on the cell grid.
  """

  # Calculate I_face based on cell grid j_toroidal
  # Equivalent to trapezoidal integration of j_toroidal_face
  I_face = jnp.concatenate([
      jnp.array([0.0]),
      math_utils.cumulative_area_integration(j_toroidal, geo),
  ])

  # Calculate d(I/F)/d(rho) at cell centers
  d_I_over_F_d_rho = jnp.diff(I_face / geo.F_face) / jnp.diff(geo.rho_face)

  return _extrapolate_to_axis(
      geo.F**3
      * geo.g3
      / (2 * jnp.pi * geo.B_0**2 * geo.rho)
      * d_I_over_F_d_rho,
      geo,
  )


def j_parallel_to_j_toroidal(
    j_parallel: array_typing.FloatVectorCell, geo: geometry.Geometry
) -> array_typing.FloatVectorCell:
  r"""Calculates j_toroidal = dI/dS from <j.B>/B0.

  Operates only on the cell grid.
  
  The relationship is .. math::

    j_\mathrm{tor} = \frac{\partial}{\partial S} \left(\frac{2 \pi B_0^2}{F}
    \int_{0}^{\rho}
        \frac{\langle j.B \rangle}{B_0 F^3 \langle 1/R^2 \rangle} \rho'
        d\rho'\right)
  
  See Eq 57
  https://users.euro-fusion.org/pages/data-cmg/wiki/files/JETTO_current.pdf

  Args:
    j_parallel: Parallel current density, <j.B>/B0, on the cell grid.
    geo: Geometry object.

  Returns:
    Toroidal current density profile on the cell grid.
  """
  # Integrand at cell centers
  integrand = j_parallel / (geo.F**3 * geo.g3) * geo.rho

  # Calculate integral
  integral = jnp.concatenate([
      jnp.array([0.0]),
      math_utils.cumulative_cell_integration(integrand * geo.rho_b, geo),
  ])

  # Calculate S_face from geo.spr
  ones = jnp.ones_like(j_parallel)
  S_face = jnp.concatenate(
      [jnp.array([0.0]), math_utils.cumulative_area_integration(ones, geo)]
  )

  # Calculate j_tor = dI/dS
  # Relationship: I = (2 pi B0^2) * (integral * F)
  I_face = (2 * jnp.pi * geo.B_0**2) * (integral * geo.F_face)

  # Calculate result on cell centers
  j_tor = jnp.diff(I_face) / jnp.diff(S_face)

  return _extrapolate_to_axis(j_tor, geo)
```

Which provides nice results for various resolutions: n_rho=250 below

image

theo-brown:
#### Effects on tests

iterhybrid-based cases (test_iterhybrid_predictor_corrector shown here)
- Minor differences in overall trajectory, such as <1% difference in Ibs
- Difference in on-axis behaviour of j profiles
- Transient differences in chi profiles that do not affect trajectory
Figure_1

circular geometry cases (test_all_transport_fusion_qlknn shown here)
- Significant differences in Ibs (~40%)
- ~10% difference in fusion gain
- Up to ~20% difference in q profile
Figure_1

I think these are all broadly expected changes, so I will update the test files.

theo-brown:
On closer inspection, it looks like the extrapolation for the on-axis values is causing some inconsistency. Turning off bootstrap current makes this a bit clearer.

test_psi_and_heat (1 is with j parallel conversions, 2 is from main)
Figure_1
Close-up on axis:
image

Will investigate iab. Initial guess it that this is probably specific to the ohmic current and whether what it's computed from matches j_total (which is toroidal).

jcitrin:
> Significant differences in Ibs (~40%)

Yeah I think that's due to bad assumptions we are making in the circular geometry model, such as constant F. This then isn't a consistent equilibrium and this could "mess up" the conversions.

Since it's just a toy model and not meant to be used for anything production, it's fine to just replace the nc files,

jcitrin:
> On closer inspection, it looks like the extrapolation for the on-axis values is causing some inconsistency. Turning off bootstrap current makes this a bit clearer.

Indeed, good point. I think it's because j_total does extrapolation to the axis term from the bulk term in `psi_calculations.calc_j_total`, but j_bootstrap goes through the parallel-->toroidal convertor with the clipping for rho_norm < 0.05.

So these should be made consistent. I actually prefer the idea of doing "something" below a certain rho_norm, since just extrapolating to [0] from [1] is fragile if there is a very fine grid, and  idx [1] may also be numerically questionable.

Perhaps we can do a combination, and do some kind of extrapolation (e.g. cubic spline) below a _RHO_MIN constant defined e.g. in psi_calculations within both the parallel<-->toroidal convertors and calc_j_total, and whose value should depend on a bit of investigation with different grid sizes.

If you agree with this approach, then I suggest not doing it in this PR. We can leave things as are now, with the slight inconsistency, and then fix it in a follow-up focused PR (adding a TODO in calc_j_total)

theo-brown:
I think that makes sense! Also makes sense to do in another PR as there isn't a spline interpolation function in the native JAX library, so there are specific design choices required there. If you're happy, I will add the comment, push the updated tests, and squash?

theo-brown:
Actually, should I switch the calc_j_total one to use the same _extrapolate_to_axis function? Then we can just update that function to use splines or whatever in the follow-up PR.

theo-brown:
@jcitrin ready for internal review. 
Haven't exhaustively checked all the test results make sense, just looked at a representative sample.
Can also squash to make a single commit if desired.

jcitrin:
Great stuff. Yes please squash the commits, I'll then bring it in

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,50 @@
...
+
+
+def _extrapolate_cell_profile_to_axis(
+    profile: array_typing.FloatVectorCell,
+    geo: geometry.Geometry,
+) -> array_typing.FloatVectorCell:
+  """Extrapolates the value of a cell profile towards the axis."""
+  # TODO: Replace with a more sophisticated extrapolation method (eg splines)
+  # Use searchsorted rather than interp, as on-axis values may be untrustworthy
+  i = jnp.searchsorted(geo.rho_norm, _MIN_RHO_NORM)
```
jcitrin: Just realized that this should probably have a min value of 1, to ensure that it's at least the first non-axis index. For edge cases where rho_norm[1]>_MIN_RHO_NORM this is needed.

Edit: actually for the cell profile it's OK since min(geo.rho_norm)>0 . But some care is maybe needed for the face profile extrapolation?

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,50 @@
...
+
+def _extrapolate_face_profile_to_axis(
+    profile: array_typing.FloatVectorFace,
+    geo: geometry.Geometry,
+) -> array_typing.FloatVectorFace:
+  """Extrapolates the value of a face profile towards the axis."""
+  # TODO: Replace with a more sophisticated extrapolation method (eg splines)
+  # Use the cell grid to find the index for extrapolation
+  # Use searchsorted rather than interp, as on-axis values may be untrustworthy
+  i = jnp.searchsorted(geo.rho_norm, _MIN_RHO_NORM)
```
jcitrin: What happens when i=0?

We don't want clamp_value = 0.5*(profile[0] + profile[i+1]) since profile[0] is the pathological one on-axis that we want to replace.

Edit: i cannot be zero for any _MIN_RHO_NORM>0. Still thinking if there's any issue for a coarse grid. Will update again momentarily

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,50 @@
...
+) -> array_typing.FloatVectorCell:
+  """Extrapolates the value of a cell profile towards the axis."""
+  # TODO: Replace with a more sophisticated extrapolation method (eg splines)
+  # Use searchsorted rather than interp, as on-axis values may be untrustworthy
+  i = jnp.searchsorted(geo.rho_norm, _MIN_RHO_NORM)
+  clamp_value = profile[i]
+  return jnp.where(geo.rho_norm < _MIN_RHO_NORM, clamp_value, profile)
+
+
+def _extrapolate_face_profile_to_axis(
```
jcitrin, line 82: OK I looked into it, and I think a small fix is needed. There are 3 cases that all need to be handled:

1. _MIN_RHO_NORM has a cell point to the left, and a face point to the right

2. _MIN_RHO_NORM has a face point to the left, and a cell point to the right

3. Special case of case 2 where the face point to the left is the axis point

For case 1, the current solution is correct.

For case 2, the current solution is also correct because the i is from the cell grid (I just noticed that...)

For case 3, then you'd want clamp_value from cell_value[0] . Maybe that should be brought into the function to be used if needed.

I also recommend unit tests for these two extrapolation functions, particularly testing all 3 cases listed above for the face_profile function.

theo-brown: Cool, I've done this including the unit test.

torax/_src/physics/psi_calculations.py
```diff
@@ -112,15 +170,14 @@ def calc_j_total(
...
-  j_total_bulk = dI_drhon[1:] / geo.spr[1:]
-  j_total_face_bulk = dI_drhon_face[1:] / geo.spr_face[1:]
-
-  # Extrapolate the axis term from the bulk term due to strong sensitivities
-  # of near-axis numerical derivatives. Set zero boundary condition on-axis
-  j_total_axis = j_total_bulk[0] - (j_total_bulk[1] - j_total_bulk[0])
+  # Note: these will be nan on axis due to division by zero, but this is
+  # handled by extrapolation below.
+  j_total = dI_drhon / geo.spr
+  j_total_face = dI_drhon_face / geo.spr_face
```
jcitrin: should never trigger NaNs, it can cause surprising behaviour like break jax gradients , even if you replace the value afterwards in the computation. Recommend putting a division guard or doing a "bulk" + "axis" pattern where the axis value is just j_total[0] from the cell grid

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,79 @@
...
+      geo: Tokamak geometry.
+
+  Returns:
+      The cell profile with modified near-axis values.
+  """
+  # TODO: Replace with a more sophisticated extrapolation method (eg splines)
+  # Current method: Extend the value of the first cell where rho_norm >=
+  # min_rho_norm to the axis.
+  mask = geo.rho_norm >= min_rho_norm
+  i = jnp.argmax(mask)
```
jcitrin, line 73: Can add a comment like:

# argmax returns the index of the first True; returns 0 if all False

Because it's a good pattern but a bit semantically weird to use argmax for the first True

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,79 @@
...
+  # cell_profile is a JitTracer, so we can index it with another JitTracer (i).
+  clamp_profile_value = cell_profile[i]
+  # geo.rho_norm is a jnp.array, so we need to explicitly dynamically index it.
+  clamp_rho_norm_value = jax.lax.dynamic_index_in_dim(geo.rho_norm, i, axis=-1)
```
jcitrin: I think can skip the `dynamic_index_in_dim` line, and just directly do:

```
return jnp.where(
    ~mask,
    clamp_profile_value,
    cell_profile
)
```

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,79 @@
...
+
+  Returns:
+      The cell profile with modified near-axis values.
+  """
+  # TODO: Replace with a more sophisticated extrapolation method (eg splines)
+  # Current method: Extend the value of the first cell where rho_norm >=
+  # min_rho_norm to the axis.
+  mask = geo.rho_norm >= min_rho_norm
+  i = jnp.argmax(mask)
+  # cell_profile is a JitTracer, so we can index it with another JitTracer (i).
```
jcitrin: no need for this comment

torax/_src/physics/psi_calculations.py
```diff
@@ -26,21 +26,79 @@
...
+  # Current method: Extend the value of the first cell where rho_norm >=
+  # min_rho_norm to the axis. Note that if there is a face i s.t.
+  # min_rho_norm <= rho_face_norm[i] <= first_cell_greater_than_min_rho_norm,
+  # the value at this face will also be modified.
+  mask = geo.rho_norm >= min_rho_norm
+  i = jnp.argmax(mask)
+  # cell_profile is a JitTracer, so we can index it with another JitTracer (i).
+  clamp_profile_value = cell_profile[i]
+  # geo.rho_norm is a jnp.array, so we need to explicitly dynamically index it.
+  clamp_rho_norm_value = jax.lax.dynamic_index_in_dim(geo.rho_norm, i, axis=-1)
```
jcitrin, line 100: here I see that indeed it is needed since mask is on a different grid from face_profile

theo-brown:
@jcitrin thanks for the review comments :)

Current density profile plots on STEP for reference: you can see that the js now do all add up to the j_total. Bonus is that j_ohmic is now visibly near zero for most of the core, as expected, so this has knock-on improvements for simulations with vloop BC.

On-axis behaviour should be investigated in a follow-up PR.

image

torax/_src/physics/psi_calculations.py
```diff
@@ -112,16 +175,21 @@ def calc_j_total(
   dI_drhon_face = jnp.gradient(Ip_profile_face, geo.rho_face_norm)
   dI_drhon = jnp.gradient(Ip_profile, geo.rho_norm)
 
+  # Note: On-axis values will be overwritten by extrapolation below, but we need
+  # to avoid division by zero
   j_total_bulk = dI_drhon[1:] / geo.spr[1:]
   j_total_face_bulk = dI_drhon_face[1:] / geo.spr_face[1:]
+  j_total_axis = jnp.array([j_total_bulk[0]])
+  j_total = jnp.concatenate([j_total_axis, j_total_bulk])
+  j_total_face = jnp.concatenate([j_total_axis, j_total_face_bulk])
```
jcitrin: The bulk-axis separation is only needed for face, where we need to avoid division by zero.

For the cell profile, can just set j_total = dI_rhon/geo.spr . We also avoid being overexuberant with the flattening if we have a coarse grid and we end up setting cell_idx[1] as the on-axis value instead of the potentially fine cell_idx[0]. If cell_idx[0] is not fine due to many grid points, then we're saved by min_rho_norm at the end.

theo-brown: Ah yes, makes sense!

PR #1502, opened 2025-09-04
Fix log_iterations in Newton-Raphson solver

theo-brown:
When running with `log_iterations=True` I was getting the following error:
```
File ".../torax/torax/_src/fvm/jax_root_finding.py", line 172, in _solver_body_logger
    jax.debug.print(
    ~~~~~~~~~~~~~~~^
        'Iteration: %d. Residual: %.16f. tau = %.6f',
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        delta_reduction,
        ^^^^^^^^^^^^^^^^
    )
    ^
  File ".../jax/_src/debugging.py", line 408, in debug_print
    formatter.format(fmt, *args, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.13/string.py", line 190, in format
    return self.vformat(format_string, args, kwargs)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.13/string.py", line 195, in vformat
    self.check_unused_args(used_args, args, kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.13/site-packages/jax/_src/debugging.py", line 355, in check_unused_args
    raise ValueError(
        f"Unused positional arguments to `jax.debug.print`: {unused_args}")
ValueError: Unused positional arguments to `jax.debug.print`: [Traced<float64[]>with<DynamicJaxprTrace>, Traced<float64[]>with<DynamicJaxprTrace>, Traced<float64[]>with<DynamicJaxprTrace>]

```

I also noticed that `_solver_body_logger` is only called in one place, with one signature (and so its external branching based on which arguments are passed is unnecessary). I've replaced the function with a direct call to `jax.debug.print`.

### DIFF
--- a/torax/_src/fvm/jax_root_finding.py
+++ b/torax/_src/fvm/jax_root_finding.py
@@ -146,39 +146,6 @@ def root_newton_raphson(
   return x_out, RootMetadata(**metadata, error=error)
 
 
-def _solver_body_logger(
-    residual: jax.Array,
-    iterations: jax.Array,
-    delta_reduction: jax.Array | None = None,
-    dt: jax.Array | None = None,
-) -> None:
-  """Logs info on internal Newton-Raphson iterations.
-
-  Args:
-    residual: Scalar residual.
-    iterations: Number of iterations taken so far in the solve block.
-    delta_reduction: Current tau used in this iteration.
-    dt: Current dt used in this iteration.
-  """
-  if dt is not None:
-    jax.debug.print(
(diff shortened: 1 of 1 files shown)


jcitrin:
@sbodenstein can you review this?

PR #1506, opened 2025-09-04
Add COCOS conversion for EQDSK

theo-brown:


theo-brown:
Revisiting the EQDSK format, I think that potentially the eqdsk files in data/third_party/geo are actually COCOS=1 rather than COCOS=2. Do correct me if I'm wrong! I found that reading them in using
```python
eq = EQDSKInterface.from_file(
    "torax/data/third_party/geo/eqdsk_cocos02.eqdsk", from_cocos=2, to_cocos=11
)
```
gives the wrong signs for COCOS=11, which should have B positive, Ip positive, psi increasing from magnetic axis to LCFS.

```
* bcentre: -1.436964362 [T] (vacuum toroidal magnetic field at the reference radius)
* cplasma: -175304.1406 [A] (plasma current)
* psibdry: -0.0 [V.s/rad or V.s] (poloidal flux at the plasma boundary (LCFS))
* psimag: 0.004152513143323447 [V.s/rad or V.s] (poloidal flux at the magnetic axis)
```
This may be because the official G-EQDSK format specifies an (R, ϕ, Z) coordinate system, which is compatible with COCOS=(1/11, 3/13, 5/15, 7/17) and incompatible with COCOS=(2/12, 4/14, 6/16, 8/18).

That said, the tests don't seem to like it. Will investigate next week.

jcitrin:
Thanks for this! This has been recently flagged to be an issue. @aaronkho 

> potentially the eqdsk files in data/third_party/geo are actually COCOS=1 rather than COCOS=2

You may be right, I'm not sure. I need to look closer myself next week. Another general issue is that I think that not all EQDSKs are necessarily in any consistent COCOS convention. 

At the moment we load with the `no_cocos=True` flag
https://github.com/google-deepmind/torax/blob/33991fe58a1d4329526b39aeac144cc182686bb4/torax/_src/geometry/geometry_loader.py#L83

If we try to load our data/third_party/geo EQDSK files with `no_cocos=False`, it seems that the eqdsk python library does not successfully determine a COCOS

image

This is quite suspicious. I'm thinking that the EQDSK files we have are not necessarily a standard COCOS? @fusionby2030 do you recall what was done here?

jcitrin:
In general for this PR, I'm thinking that maybe instead of having options internally to convert to any COCOS, we should do the following:

1. Actually finally set what the TORAX COCOS is going to be
2. Make all the internal changes to be consistent with that COCOS
3. Then change the EQDSK reader to convert from whatever the EQDSK is, to the TORAX COCOS. If the reader cannot successfully determine the COCOS, then we raise an Exception and the onus is on the user to provide valid geometry data.

theo-brown:
> Another general issue is that I think that not all EQDSKs are necessarily in any consistent COCOS convention.

Someone with more EQDSK experience can correct me if I'm wrong, but I believe that it is not possible to fully identify a COCOS from an EQDSK without more information.

From the [`eqdsk` package source code](https://github.com/Fusion-Power-Plant-Framework/eqdsk/blob/e571f6602126108a032d08e47e3a26b1bb170734/eqdsk/cocos.py#L312-L315):
```python
def identify_eqdsk(
    eqdsk: EQDSKInterface,
    *,
    clockwise_phi: bool | None = None,
    volt_seconds_per_radian: bool | None = None,
    qpsi_positive: bool | None = None,
) -> list[COCOS]:
    ...
    cw_phi_l = [True, False] if clockwise_phi is None else [clockwise_phi]
    vs_pr_l = (
        [True, False] if volt_seconds_per_radian is None else [volt_seconds_per_radian]
    )
    ...
```
That is, it's not possible to distinguish whether a provided EQDSK has:
- ψ in Vs/rad or Vs (ie ψ/2π vs ψ, and hence COCOS X vs COCOS 1X)
- Φ clockwise or anticlockwise (ie even COCOS vs odd COCOS)

This is why the COCOS of the EQDSKs in data/third_party/geo can't be fully identified and could be any in 1, 11, 2, 12.

> In general for this PR, I'm thinking that maybe instead of having options internally to convert to any COCOS, we should do the following:

    1. Actually finally set what the TORAX COCOS is going to be
    2. Make all the internal changes to be consistent with that COCOS
    3. Then change the EQDSK reader to convert from whatever the EQDSK is, to the TORAX COCOS. If the reader cannot successfully determine the COCOS, then we raise an Exception and the onus is on the user to provide valid geometry data.

Partial agree, only disagree is with 3. In this PR I have:

1. Assumed TORAX is COCOS 11. This is what the EQDSK convertor was doing silently anyway. The compatible COCOS with TORAX in its present state are COCOS 11 and COCOS 12:

| | COCOS 11 | COCOS 12 | 
|-|----------------|-----------------|
| psi | increasing | increasing |
| Ip | positive | positive |
| B_tor | positive | positive |
| q | positive | positive |
| ψ_cocos = ψ/2π? | no | no |
|  Φ clockwise from top? | no | yes |

2. I think until rotation / momentum transport is implemented there is no practical difference between 11 and 12. TORAX is currently compatible with both.

3. Set the EQDSK loader to explicitly (rather than silently) convert to COCOS 11. The onus is on the user is *to explicitly set the COCOS of the EQDSK in config* (supports any COCOS). I think this is the most robust and transparent option.

theo-brown:
It would be useful to know how the EQDSK files in data/third_party/geo were generated, so that we can check some of this. E.g. were they silently converted to odd COCOS when they were written?

theo-brown:
The difference in the tests is 2\pi, so it looks like these EQDSKs are actually already COCOS11.

However, I think one of them maybe has an additional factor of 2pi floating around - in iterhybrid_predictor_corrector.eqdsk, I get the following difference in the psi profiles

image

Which equates exactly to a difference of 2pi * eqfile['psimag'].

theo-brown:
@fusionby2030 could I ask you some questions about this? Could be either over email, on a call, or in this PR discussion.

@jcitrin fyi it's quite useful to have this sorted as I think it fixes some problems with loading EQDSK from JETTO.

fusionby2030:
Sure, email or in the PR.

theo-brown:
@fusionby2030 what did you use to generate these EQDSKs?

torax/_src/geometry/standard_geometry.py
```diff
@@ -708,9 +712,38 @@ def calculate_area(x, z):
...
+        cocos,
     )
+    COCOS_violations = ''
+    if eqfile['bcentre'] < 0:
+      COCOS_violations += '- B_0 is negative\n'
+    if eqfile['psibdry'] < eqfile['psimag']:
+      COCOS_violations += '- psi at the boundary is less than psi on the axis\n'
+    if np.any(eqfile['fpol']) < 0:
+      COCOS_violations += '- F=RB_phi has negative values\n'
+    if np.any(eqfile['fpol']) < 0:
```
jcitrin: Typo (repeated lines)

torax/_src/geometry/standard_geometry.py
```diff
@@ -985,7 +1018,7 @@ def calculate_area(x, z):
         a_minor=a_minor,
         B_0=np.array(B_0),
         # TODO(b/335204606): handle COCOS shenanigans
```
jcitrin: can remove this TODO?

torax/_src/geometry/standard_geometry.py
```diff
@@ -708,9 +712,38 @@ def calculate_area(x, z):
...
     # ---- 1. Load the eqdsk ---- #
     # --------------------------- #
+    if cocos in [2, 4, 6, 8, 12, 14, 16, 18]:
+      logging.warning(
+          f'Requested COCOS {cocos} has a (R, Z, phi) coordinate system'
+          ' (sigma_RphiZ = -1), but the EQDSK format specifies a (R, phi, Z)'
+          ' coordinate system (sigma_RphiZ = +1). This may result in'
+          ' unexpected behaviour.'
+      )
     eqfile = geometry_loader.load_geo_data(
```
jcitrin, line 795: Add a comment that the geometry loader does the conversion to COCOS 11

torax/_src/geometry/standard_geometry.py
```diff
@@ -708,9 +712,38 @@ def calculate_area(x, z):
...
+    if np.any(eqfile['qpsi']) < 0:
+      COCOS_violations += '- q has negative values\n'
+    if eqfile['cplasma'] < 0:
+      COCOS_violations += '- Ip is negative'
+    if COCOS_violations:
+      raise ValueError(
+          'The following violate the COCOS11 coordinate system:\n'
+          + COCOS_violations
+          + 'Check that the COCOS of the input EQDSK was set correctly.'
+      )
```
jcitrin: Nit: the COCOS 11 validation can be put into a new helper private function

torax/_src/geometry/pydantic_model.py
```diff
@@ -241,6 +244,7 @@ class EQDSKConfig(torax_pydantic.BaseModelFrozen):
   geometry_file: str = 'EQDSK_ITERhybrid_COCOS02.eqdsk'
   n_surfaces: pydantic.PositiveInt = 100
   last_surface_factor: torax_pydantic.OpenUnitInterval = 0.99
+  cocos: pydantic_types.COCOSInt
```
jcitrin: COCOSInt can be defined here in geometry.pydantic_model

jcitrin:
Sorry for delays in responding. If you're convinced that the EQDSK in data/third_party/geo are in an inconsistent COCOS state, is it possible to modify directly the offending values to be consistent with something the convertor can handle, and we are then able to move on?

torax/_src/geometry/geometry_loader.py
```diff
@@ -79,15 +80,11 @@ def _load_fbt_data(file_path: str | IO[bytes]) -> dict[str, np.ndarray]:
...
-  eqdsk_data = eqdsk.EQDSKInterface.from_file(file_path, no_cocos=True)
-  # TODO(b/335204606): handle COCOS shenanigans
-  logging.warning(
-      """
-               WARNING: Using EQDSK geometry.
-               The TORAX EQDSK converter has only been tested against CHEASE-generated EQDSK which is COCOS=2.
-               The converter is not guaranteed to work as expected with arbitrary EQDSK input. Please verify carefully.
-               Future work will be done to correctly handle EQDSK inputs provided with a specific COCOS value."""
+def _load_eqdsk_data(
+    file_path: str, cocos: pydantic_types.COCOSInt
```
jcitrin: typehint update

torax/_src/geometry/geometry_loader.py
```diff
@@ -103,8 +100,11 @@ def load_geo_data(
     geometry_dir: str | None,
     geometry_file: str,
     geometry_source: GeometrySource,
+    cocos: pydantic_types.COCOSInt | None = None,
```
jcitrin: typehint update

theo-brown:
> If you're convinced that the EQDSK in data/third_party/geo are in an inconsistent COCOS state, is it possible to modify directly the offending values to be consistent with something the convertor can handle, and we are then able to move on?

Can do! In order to keep track of provenance, could you help me understand what the difference is between `EQDSK_ITERhybrid_COCOS02.eqdsk` and `eqdsk_cocos02.eqdsk`, and where they came from? I wasn't quite clear from the README, sorry if I'm being dense.

EDIT: Conclusion - `EQDSK_ITERhybrid_COCOS02.eqdsk` is directly from [PINT](https://gitlab.com/qualikiz-group/pyntegrated_model/-/tree/main/geo?ref_type=heads), `eqdsk_cocos02.eqdsk` is unknown and has been removed.

theo-brown:
Any tips on breaking the cyclic import problem? I know there are a few different options (type hinting with a string, using from __future__, etc) but I don't know which is preferred under the Google style guide.

theo-brown:
The two failing unit tests are due to 2pi differences in the psi profiles. I think this is a bug in the Python eqdsk package.

theo-brown:
2π bug has now been fixed in eqdsk>=0.7.0 (thanks @je-cook!).

theo-brown:
Comparison of modified geometry loaders (this branch) with original geometry loaders (unmodified test_iterhybrid_predictor_corrector.nc):
image

Note this isn't quite apples-to-apples as I am comparing just the geometry loaders with the geometry variables from test_iterhybrid_predictor_corrector.nc, which has renormalisation of the Ip profile to achieve the target Ip. Hence, we expect the reference to differ slightly in psi, Ip and j_total.

theo-brown:
EQDSK geometry now has psi[axis]=0. This is more consistent with the CHEASE geometry loader. I've updated the test file to reflect this.

Before (solid), after (dashed)
image

New EQDSK  vs CHEASE
image

theo-brown:
@jcitrin ready for re-review. Scope ended up creeping a bit!

jcitrin:
@theo-brown sorry for taking so long. Planned to review and bring in this week. Could you sync to head again?

jcitrin:
> EQDSK geometry now has psi[axis]=0. This is more consistent with the CHEASE geometry loader. I've updated the test file to reflect this.

This was a deliberate choice.

https://github.com/google-deepmind/torax/pull/1487

The reason is that EQDSK files could be generated from other workflows where the absolute value of psi does have meaning, e.g. flux consumption. I will suggest to revert in the review

jcitrin:
Sorry again for the delays. I'm reviewing now. Could you please sync (yet again) in the meantime?

torax/_src/geometry/standard_geometry.py
```diff
@@ -768,21 +777,28 @@ def from_eqdsk(
...
-      area = 0.0
-      for i in range(n):
-        j = (i + 1) % n  # roll over at n
-        area += x[i] * z[j]
-        area -= z[i] * x[j]
-      area = abs(area) / 2.0
-      return area
+      z_shifted = np.roll(z, -1)
+      x_shifted = np.roll(x, -1)
+      return 0.5 * np.abs(np.sum(x * z_shifted - z * x_shifted))
```
jcitrin, line 782: Nice find

torax/_src/geometry/standard_geometry.py
```diff
@@ -1056,23 +1066,22 @@ def calculate_area(x, z):
         R_major=R_major,
         a_minor=a_minor,
         B_0=np.array(B_0),
-        # TODO(b/335204606): handle COCOS shenanigans
-        psi=(psi_on_flux_surfaces + eqfile['psimag']) * 2 * np.pi,
-        Ip_profile=Ip_eqdsk,
-        Phi=Phi_eqdsk,
+        psi=psi_on_flux_surfaces,
```
jcitrin: should still add psimag since depending on the application it could have meaning (e.g. related to flux consumption)

torax/examples/step_flattop_bgb.py
```diff
@@ -79,19 +83,17 @@
...
-        }
-    ),
+    "profile_conditions": profile_conditions_from_ids | {
+        "use_v_loop_lcfs_boundary_condition": True,
+        "v_loop_lcfs": 0.0,
+    },
     "geometry": {
         # TODO(b/323504363): Switch to loading from IMAS rather than eqdsk
         # Currently there is a bug in either the TORAX IMAS loader or the JETTO
         # IMAS writer that makes them incompatible
```
jcitrin, line 93: is this still the case? Note that there has been a recent small fix to the TORAX IMAS loader if you sync, where <1/R> is correctly used for vpr --> spr . Not sure if this is related

theo-brown: Thanks for suggesting I look into this. It's exposed a few other problems that I'll have a look at next week!

torax/_src/geometry/geometry_loader.py
```diff
@@ -103,8 +114,11 @@ def load_geo_data(
     geometry_dir: str | None,
     geometry_file: str,
     geometry_source: GeometrySource,
+    cocos: COCOSInt | None = None,
 ) -> dict[str, np.ndarray]:
-  """Loads the data from a geometry file into a dictionary."""
+  """Loads the data from a geometry file into a dictionary.
+
+  If geometry_source=EQDSK, cocos must be provided."""
```
jcitrin, line 121: @theo-brown quick question that came up in internal review.

Is this accurate, or if not provided, doesn't the eqdsk  library try to identify the convention and raise an error if it is not possible?

No need to make any new commits here, we can do such nits internally.

theo-brown: COCOS are not uniquely identifiable from data in an EQDSK. As an example, a major part of COCOS conventions is whether psi is normalised by 2pi or not. This information is not stored in the EQDSK format - you just get given an array of psi values without information as to whether they are in units of radians or not.

This is one of the reasons EQDSK is a legacy format!

theo-brown:
> One issue that came up internally is the new folder "third_party" naming breaks some directory naming conventions. Therefore will just move everything there back to data/imas (not data/imas_data which is a redundant name).

There's already data/third_party/geo, why is data/third_party/imas not acceptable? Or is it that you can't have two subdirectories within third_party (weird)?

jcitrin:
> There's already data/third_party/geo, why is data/third_party/imas not acceptable? Or is it that you can't have two subdirectories within third_party (weird)?

Ah, now I understand what's going on. 

1. The naming of data/third_party/geo is actually a transformation of the internal directory structure `data/geo` to the external `data/third_party/geo` . The basic reason is that there are rules around the usage of `third_party`.

2. imas_data never went through that transformation. It should. I will do so now, so externally it will be seen as `data/third_party/imas` as you did here. Because there was no copybara rule yet for the `imas_data` (now `imas`) transformation, the import of your new `data/third_party/imas` led to an invalid internal directory structure.

tl;dr . I will fix this internally and you will see `data/third_party/imas` as expected

theo-brown:
Nice, thanks for looking into it :)

torax/examples/step_flattop_bgb.py
theo-brown: We think the JETTO equilibrium IDS might be inconsistent with the EQDSK. This is related to something that Emmi is looking into more widely in the JINTRAC source repo. Once we've pinned down the issue I will revisit this todo. For the time being, we can leave this as loading from eqdsk, if that's ok? I don't think it's vital to this PR.

jcitrin:
Almost there with merging this.

A point came up in internal review on the change in `standard_geometry_test`

```
# EQDSK is more inaccurate
np.testing.assert_allclose(eqdsk_geo.gm4, chease_geo.gm4, rtol=0.05)
```

where rtol increased to 0.05 from 0.01.

What originally motivated this? None of the changes to the EQDSK convertor should have impacted accuracy, and when we revert the changes back to the original tolerance, the tests pass.

No need to make any code reversions, we are doing it internally, but just wondering what triggered the original modification?

theo-brown:
> What originally motivated this? None of the changes to the EQDSK convertor should have impacted accuracy, and when we revert the changes back to the original tolerance, the tests pass.
> No need to make any code reversions, we are doing it internally, but just wondering what triggered the original modification?

Locally, setting rtol=0.01 causes it to fail with

```
AssertionError: 
E     Not equal to tolerance rtol=0.01, atol=0
E     
E     Mismatched elements: 1 / 25 (4%)
E     Max absolute difference among violations: 0.38738439
E     Max relative difference among violations: 0.0107102
```

jcitrin:
There were 2 EQDSK checks, originally with

```
np.testing.assert_allclose(imas_geo.gm4, chease_geo.gm4, rtol=0.01)
np.testing.assert_allclose(eqdsk_geo.gm5, chease_geo.gm5, rtol=0.02)
```

This passes locally for me. Above did you set 0.01 for the gm5 check?

theo-brown:
Ah you're right the original value for gm5 was 0.02, not 0.01. I think the change was originally a mistake from just copying the wrong line when playing around with the tests. Sorry!

jcitrin:
np! Merging now

PR #1535, opened 2025-09-08
Use exact <B^2> and <1/B^2> in Angioni-Sauter

theo-brown:
Use the new <B^2> and <1/B^2> in Angioni Sauter model.


theo-brown:
Almost no changes to iterhybrid_predictor_corrector_neoclassical.nc, just had to update to pass the test:

angioni_sauter

theo-brown:
Changes to reference values in angioni_sauter_test.py are ~2% in the core and ~25% at the edge for chi, D, V, and ~1e-6% for V_ware_e.

theo-brown:
Differences in an ST case are slightly more pronounced, but not game-changing. (1), solid, is using the approximation from circular geometry, (2), dashed is using the proper calculation. chi_i still missing on-axis behaviour we see in NCLASS (and so potato fudge is still on the menu).

st

Also, we're internally in the process of getting approval for publicly releasing an ST prototype power plant design (including geometry and profiles data) so hopefully these kind of checks can be incorporated into the test suite at some point in the not-too-distant future.

Issue #1572, opened 2025-09-18
Pytest failures for fresh install of TORAX


tbody-cfs:
OS: Mac 15.7
CPU: Apple M3 Pro
Python version: 3.13.5
Git hash: a4dbe317867c86db4ed660fea2e178e35c7e3078

Fresh install in virtual environment with `pip install -e . [dev]`. Versions of installed packages in [versions.txt](https://github.com/user-attachments/files/22407519/versions.txt).

50 tests failed. Short test summary is
```
FAILED torax/_src/tests/array_typing_test.py::ArrayTypingTest::test_dataclass - AssertionError: TypeCheckError not raised
FAILED torax/_src/tests/array_typing_test.py::ArrayTypingTest::test_invalid_shape - AssertionError: TypeCheckError not raised
FAILED torax/_src/tests/interpolated_param_test.py::InterpolatedParamTest::test_interpolated_param_is_invariant_to_xs_order0 - AssertionError: RuntimeError not raised
FAILED torax/_src/tests/interpolated_param_test.py::InterpolatedParamTest::test_interpolated_param_is_invariant_to_xs_order1 - AssertionError: RuntimeError not raised
FAILED torax/_src/tests/jax_utils_test.py::JaxUtilsTest::test_enable_errors - AssertionError: RuntimeError not raised
FAILED torax/tests/sim_experimental_compile_test.py::SimExperimentalCompileTest::<<all>> - jax.errors.TracerArrayConversionError: The numpy.ndarray conversion method __array__() was called on traced array with shape float64[]
FAILED torax/tests/sim_time_dependence_test.py::SimWithTimeDependenceTest::test_time_dependent_params_update_in_adaptive_dt_with_adaptive_dt - TypeError: Value 'fake' for discriminator 'model_name' mapped to multiple choices
FAILED torax/tests/sim_time_dependence_test.py::SimWithTimeDependenceTest::test_time_dependent_params_update_in_adaptive_dt_without_adaptive_dt - TypeError: Value 'fake' for discriminator 'model_name' mapped to multiple choices
```

Full pytest trace: [pytest.txt](https://github.com/user-attachments/files/22407902/pytest.txt)

Nush395:
Hey Tom,

Thanks for raising the issue. There's a few different problems going on here.

[...]

PR #1600, opened 2025-09-25
Document required environment variables for pytest in README

yadneshx17:
This PR addresses the first point of [this comment](https://github.com/google-deepmind/torax/issues/1572#issuecomment-3311477175)

- Adds a section under "Set environment variables" in README.md.
- Informs contributors to set the following environment variables when running pytest:
  - TORAX_JAXTYPING=True
  - TORAX_ERRORS_ENABLED=True
- Ensures that all assertion and error checks are active during testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,15 @@ run_torax  --config='examples/iterhybrid_rampup.py' --log_progress
 
 #### Set environment variables
 
+##### Running Tests with Full Assertion Checks
+
+To run the pytest test suite with all assertion checks enabled, set the following environment variables:
+
+```shell
+$ export TORAX_JAXTYPING=True
+$ export TORAX_ERRORS_ENABLED=True
+```
+
 ##### Error checking
 
 If true, error checking is enabled in internal routines. Used for debugging.


README.md
```diff
@@ -151,6 +151,15 @@ run_torax  --config='examples/iterhybrid_rampup.py' --log_progress
 
 #### Set environment variables
 
+##### Running Tests with Full Assertion Checks
+
+To run the pytest test suite with all assertion checks enabled, set the following environment variables:
+
+```shell
+$ export TORAX_JAXTYPING=True
```
Nush395: I think it would be nice to write down the actual command used to run pytest here so that users can simply copy and paste a test command.

yadneshx17: Good idea,  I’ll update the README to include the full pytest command. thanks for the suggestion!

yadneshx17: I’ve updated the PR to include the actual pytest command so users can copy and paste it. Please let me know if any further tweaks are needed.

README.md
```diff
@@ -196,6 +196,30 @@ run_torax \
...
+
+To run the pytest test suite with all assertion checks enabled, you can use the following commands depending on your platform:
+
+**Linux / macOS (bash/zsh):**
+
+```bash
+TORAX_JAXTYPING=True TORAX_ERRORS_ENABLED=True pytest . -n auto
+```
+
+**Windows:**
```
Nush395: For simplicity I think we can just document the linux command

yadneshx17: Updated the README as suggested

Nush395 (approved):
Thanks! Can you also squash your commits please?

Nush395 (changes requested):
Can you squash your commits before merging?

yadneshx17:
Done!

PR #1619, opened 2025-10-06
plasma_composition converter for imas core_profiles

MateoBell:
Followup of #1486. 
Adds function to map plasma_composition data from core_profiles IDS, 
Also creates a global function which writes everything useful as TORAX input from core_profiles IDS. 

Always map Z_eff and all impurity fractions as ne ratios. 
Intended use: When dumped into a config object, manually set the impurity mode between n_e_ratios and n_e_ratios_Z_eff. If set to n_e_ratios_Z_eff, should also override one impurity ratio to None.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -2092,10 +2092,11 @@ saved in Data Dictionary version 4.0.0 or newer.
 If the IDS is stored in an IMAS db or in a netCDF file it can be loaded using
 the loader function ``load_imas_data`` from |imas_loader|.
 It can then be loaded programatically in the ``CONFIG`` by constructing a nested
-dictionary with the ``core_profiles_from_imas`` function from
-|core_profiles_input_imas|. The function returns a dictionary whose structure
-fits the schema of profile_conditions and can be programatically loaded into a
-``CONFIG`` with standard dictionary manipulation.
+dictionary with the ``profile_conditions_from_imas`` and
+``plasma_composition_from_imas`` functions from |core_profiles_input_imas|. The
+functions returns a dictionary whose structure fits the schema of 
+profile_conditions or plasma_composition and can be programatically loaded into
+a ``CONFIG`` with standard dictionary manipulation.
 
 An example on how to inject the IMAS conditions into the config can be found in
 the test file ``imas_tools/input/tests/core_profiles_test.py``.
--- a/torax/_src/constants.py
+++ b/torax/_src/constants.py
@@ -109,3 +109,5 @@ ION_PROPERTIES_DICT: Final[Mapping[str, IonProperties]] = (
 )
 
 ION_SYMBOLS = frozenset(ION_PROPERTIES_DICT.keys())
+
+HYDROGENIC_IONS: Final[frozenset[str]] = frozenset(['H', 'D', 'T'])
--- a/torax/_src/imas_tools/input/core_profiles.py
+++ b/torax/_src/imas_tools/input/core_profiles.py
@@ -12,17 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Useful functions to load IMAS core_profiles or plasma_profiles IDSs."""
-from typing import Any, Mapping
+from typing import Any, Collection
 
 from imas import ids_toplevel
 import numpy as np
+from torax._src import constants
 
 
 # pylint: disable=invalid-name
 def profile_conditions_from_IMAS(
     ids: ids_toplevel.IDSToplevel,
     t_initial: float | None = None,
-) -> Mapping[str, Any]:
+) -> dict[str, Any]:
   """Converts core_profiles IDS to a profile_conditions dict for TORAX config.
 
   Args:
@@ -36,11 +37,9 @@ def profile_conditions_from_IMAS(
     The updated fields read from the IDS that can be used to completely or
     partially fill the `profile_conditions` section of a TORAX `CONFIG`.
   """
-  profiles_1d = ids.profiles_1d
-  time_array = [profile.time for profile in profiles_1d]
-  if t_initial:
-    time_array = [ti - time_array[0] + t_initial for ti in time_array]
-  rhon_array = [profile.grid.rho_tor_norm for profile in profiles_1d]
+  profiles_1d, rhon_array, time_array = _get_time_and_radial_arrays(
+      ids, t_initial
+  )
 
   # profile_conditions
   psi = (np.array(rhon_array[0]), np.array(profiles_1d[0].grid.psi))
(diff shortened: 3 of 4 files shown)


torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +110,110 @@ def profile_conditions_from_IMAS(
       "normalize_n_e_to_nbar": False,
       "v_loop_lcfs": v_loop_lcfs,
   }
+
+
+def plasma_composition_from_imas(
```
Nush395: nit: keep the naming consistent with the other function which has upper case `IMAS`

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +110,110 @@ def profile_conditions_from_IMAS(
...
+    ids: ids_toplevel.IDSToplevel,
+    t_initial: float | None = None,
+) -> Mapping[str, Any]:
```
Nush395: In this case as we might override fields of the output dict I'd recommend an output type of `dict` or `MutableMapping` as these support `__setitem__` and `__delitem` which we might want to do on the returned object.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +110,110 @@ def profile_conditions_from_IMAS(
...
+  """Returns dict with args for plasma_composition config from a given ids.
+
+  Loading IMAS data for plasma composition should only be used with n_e_ratios
+  and n_e_ratios_Z_eff impurity modes, not with fractions mode. The impurity
+  mode needs to be specified explicitly after loading the IMAS data. In case
+  n_e_ratios_Z_eff is specified, one impurity ratio must be set to None
+  explicitly. In case n_e_ratios is used, Z_eff will simply be ignored.
+  Note that if the ids indivual ions properties are not filled, it will not
```
Nush395: typo: individual

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +110,110 @@ def profile_conditions_from_IMAS(
...
+  raise an error and just return an empty dict as main_ion and species.
```
Nush395: should we raise an error here to fail loudly and informatively?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +110,110 @@ def profile_conditions_from_IMAS(
...
+
+  Returns:
+    The updated fields read from the IDS that can be used to completely or
+    partially fill the `plasma_composition` section of a TORAX `CONFIG`.
+  """
+  profiles_1d = ids.profiles_1d
+  time_array = [profile.time for profile in profiles_1d]
+  if t_initial:
+    time_array = [ti - time_array[0] + t_initial for ti in time_array]
+  rhon_array = [profile.grid.rho_tor_norm for profile in profiles_1d]
```
Nush395: extract this into a helper function that is used by both functions

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +110,110 @@ def profile_conditions_from_IMAS(
...
+      "impurity": {
+          "species": species,
+      },
+  }
+
+
+def imas_core_profiles_converter(
+    ids: ids_toplevel.IDSToplevel,
+    t_initial: float | None = None,
+) -> Mapping[str, Mapping[str,Any]]:
```
Nush395: again return `dict` here for reasons above

torax/_src/imas_tools/input/core_profiles.py
Nush395: As a general point for these loaders as IMAS doesn't give us any guarantees on what data is filled in I'm leaning towards doing some input validation. wdyt?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +106,115 @@ def profile_conditions_from_IMAS(
...
+      time_array,
+      rhon_array,
+      [profile.zeff for profile in profiles_1d],
+  )
+  species = {}
+  ratios = {}
+  for ion in range(len(profiles_1d[0].ion)):
+    try:
+      symbol = str(profiles_1d[0].ion[ion].name)
+    except (AttributeError):
```
Nush395: can we make some assumptions/restrictions on the DD that TORAX expects to make this logic simpler?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +106,115 @@ def profile_conditions_from_IMAS(
...
+        # accurate to take ratios of volume integrated densities ?
+  total_main_ion_density = np.sum([ratio for ratio in ratios.values()], axis=0)
+  main_ion = {}
+  for symbol, ratio in ratios.items():
+    main_ion[symbol] = (time_array, ratio / total_main_ion_density)
+  return {
+      "main_ion": main_ion,
+      "Z_eff": Z_eff,
+      "impurity": {
+          "species": species,
```
Nush395: should also set `"impurity_mode": "n_e_ratios"` here

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +106,115 @@ def profile_conditions_from_IMAS(
...
+  ratios = {}
+  for ion in range(len(profiles_1d[0].ion)):
+    try:
+      symbol = str(profiles_1d[0].ion[ion].name)
+    except (AttributeError):  
+      # Case ids is plasma_profiles in early DDv4 releases.
+      symbol = str(profiles_1d[0].ion[ion].label)
+    if symbol in constants.ION_PROPERTIES_DICT.keys():
+      # Fill impurities
+      if symbol not in ("D", "T", "H"):
```
Nush395: can these main ion symbols be brought out as constants? If these will always be disambiguated from other ions as impurities they could be in `torax._src.constants`

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +106,115 @@ def profile_conditions_from_IMAS(
...
+  return {
+      "main_ion": main_ion,
+      "Z_eff": Z_eff,
+      "impurity": {
+          "species": species,
+      },
+  }
+
+
+def IMAS_core_profiles_converter(
```
Nush395: is there value in having this separate function?

I'm imagining that in practice for now users will use the `profile_conditions_from_IMAS` and `plasma_composition_from_IMAS` in a `CONFIG` themselves. Is there a separate use case for this function?

torax/_src/imas_tools/input/core_profiles.py
MateoBell: plasma_profiles was introduced from DD4.0.0 which is the latest version currently, and from DD4.1 and later it should be unified and we can simplify the logic. So if possible I would like to wait for a new minor release to simplify the logic and add a restriction on the DD to use this newer version.

torax/_src/imas_tools/input/core_profiles.py
MateoBell: ok ! I added a frozenset in `torax._src.constants`. Not entirely sure about how to deal with `H` or `He3/4` which can be treated as main ions or impurities depending on the scenario I think ?

torax/_src/constants.py
```diff
@@ -109,3 +109,5 @@ class Constants:
 )
 
 ION_SYMBOLS = frozenset(ION_PROPERTIES_DICT.keys())
+
+MAIN_ION_SYMBOLS = frozenset(['H', 'D', 'T'])
```
tamaranorman: Maybe call this HYDROGENIC_IONS to be more accurate and also can you add a type for this

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +108,90 @@ def profile_conditions_from_IMAS(
...
+  ratios = {}
+  for ion in range(len(profiles_1d[0].ion)):
+    try:
+      symbol = str(profiles_1d[0].ion[ion].name)
+    except AttributeError:
+      # Case ids is plasma_profiles in early DDv4 releases.
+      symbol = str(profiles_1d[0].ion[ion].label)
+    if symbol in constants.ION_PROPERTIES_DICT.keys():
+      # Fill impurities
+      if symbol not in constants.MAIN_ION_SYMBOLS:
```
tamaranorman: Make this an optional parameter to allow users to set if they want to allow other things to be main ions

Also maybe switch to have the positive case first to simplify the control flow

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +108,90 @@ def profile_conditions_from_IMAS(
...
+            rhon_array,
+            [
+                profile.ion[ion].density / profile.electrons.density
+                for profile in profiles_1d
+            ],
+        )
+        species[symbol] = n_e_ratio
+      # Fill main ions
+      else:
+        ratios[symbol] = [
```
tamaranorman: Maybe clean-up this name to main_ion_ratios

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +108,90 @@ def profile_conditions_from_IMAS(
...
+      if symbol not in constants.MAIN_ION_SYMBOLS:
+        n_e_ratio = (
+            time_array,
+            rhon_array,
+            [
+                profile.ion[ion].density / profile.electrons.density
+                for profile in profiles_1d
+            ],
+        )
+        species[symbol] = n_e_ratio
```
tamaranorman: Can this name also be expanded slightly

torax/_src/imas_tools/input/tests/core_profiles_test.py
```diff
@@ -107,6 +107,54 @@ def test_init_profiles_from_IMAS(self):
...
+    np.testing.assert_allclose(
+        torax_config.plasma_composition.impurity.species["Ne"].get_value(0.0),
+        0.01,
+    )
+    np.testing.assert_allclose(
+        torax_config.plasma_composition.main_ion["D"].get_value(0.0), 0.5
+    )
+    np.testing.assert_equal(
+        torax_config.plasma_composition.get_main_ion_names(), ("D", "T")
+    )
```
tamaranorman: Also add a test that the ratios etc are calculated correctly?

docs/configuration.rst
```diff
@@ -2051,10 +2051,14 @@ saved in Data Dictionary version 4.0.0 or newer.
...
-dictionary with the ``core_profiles_from_imas`` function from
-|core_profiles_input_imas|. The function returns a dictionary whose structure
-fits the schema of profile_conditions and can be programatically loaded into a
-``CONFIG`` with standard dictionary manipulation.
+dictionary with the ``profile_conditions_from_imas`` and
+``plasma_composition_from_imas`` functions from |core_profiles_input_imas|. The
+functions returns a dictionary whose structure fits the schema of 
+profile_conditions or plasma_composition and can be programatically loaded into
+a ``CONFIG`` with standard dictionary manipulation.
+In case you want to directly load both plasma_composition and
```
tamaranorman: Does this function exist yet? It doesn't seem to be here

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -12,17 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Useful functions to load IMAS core_profiles or plasma_profiles IDSs."""
-from typing import Any, Mapping
+from typing import Any
 
 from imas import ids_toplevel
 import numpy as np
+import torax._src.constants as constants
```
tamaranorman: from torax._src import constants

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +108,90 @@ def profile_conditions_from_IMAS(
...
   }
+
+
+def plasma_composition_from_IMAS(
+    ids: ids_toplevel.IDSToplevel,
+    t_initial: float | None = None,
+) -> dict[str, Any]:
+  """Returns dict with args for plasma_composition config from a given ids.
+
+  Loading IMAS data for plasma composition should only be used with n_e_ratios
```
tamaranorman: Can this maybe be simplified and tuned into bullet points - something like:

The profiles returned from this can be used with two impurity modes (contained within the `impurity` key
* `n_e_ratios` -   default returned from this function (note Z_eff will be ignored)
* `n_e_ratios_Z_eff` - one impurity should be set to None and the `impurity_mode` string set
* `fractions` - unsupported

Then more clearly add the note about not being filled

For safety is it worth taking the ions that are expected and error if we don't have values for this if it could be due to them being in the wrong places

torax/_src/imas_tools/input/tests/core_profiles_test.py
MateoBell: isnt it covered by the previous lines or do you mean something else ? `np.testing.assert_allclose(torax_config.plasma_composition.impurity.species["Ne"].get_value(0.0), 0.01,)
np.testing.assert_allclose(torax_config.plasma_composition.main_ion["D"].get_value(0.0), 0.5)`

torax/_src/imas_tools/input/core_profiles.py
MateoBell: > For safety is it worth taking the ions that are expected and error if we don't have values for this if it could be due to them being in the wrong places

Do you mean adding an extra optional arg to specify which ions are expected and then check if they are present in the IDS ?

tamaranorman: I do yes, as then it is a more explicit check - although this does mean that a user would have to have more knowledge of what they are loading but maybe that is also a good thing

torax/_src/imas_tools/input/tests/core_profiles_test.py
tamaranorman: Sorry I missed this, it feels like maybe the ordering is a bit wrong

torax/_src/imas_tools/input/tests/core_profiles_test.py
```diff
@@ -107,6 +107,54 @@ def test_init_profiles_from_IMAS(self):
...
+    # Modifying the input config profiles_conditions class
+    plasma_composition_data = core_profiles.plasma_composition_from_IMAS(
+        core_profiles_in,
+    )
+    config["plasma_composition"] = plasma_composition_data
+    config["plasma_composition"]["impurity"]["impurity_mode"] = "n_e_ratios"
+    torax_config = model_config.ToraxConfig.from_dict(config)
+
+    assert torax_config.plasma_composition.impurity.species["Ne"] is not None
```
tamaranorman: Do
ne_impurity = ....
then self.assertIsNotNone(ne_impurity)?

torax/_src/imas_tools/input/tests/core_profiles_test.py
MateoBell: indeed you are right makes more sense having them after !

torax/_src/imas_tools/input/core_profiles.py
MateoBell: Ok ! I added optional checks for both main_ions and impurities (and related subtests), let me know if the implementation is too heavy.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +108,135 @@ def profile_conditions_from_IMAS(
...
+    partially fill the `plasma_composition` section of a TORAX `CONFIG`.
+  """
+  profiles_1d, rhon_array, time_array = _get_time_and_radial_arrays(
+      ids, t_initial
+  )
+  # Check that the expected ions are present in the IDS
+  ids_ions = [ion.name for ion in profiles_1d[0].ion if ion.density]
+  if expected_impurities:
+    for impurity in expected_impurities:
+      if impurity not in constants.ION_PROPERTIES_DICT.keys():
```
tamaranorman: This and the below are pretty similar code, maybe merge into a helper function?

tamaranorman (commented):
Looks good just one small change and then we should be able to get this merged in this week, apologies for the delay

tamaranorman:
Do you want to then also squash the change and I will import it

PR #1630, opened 2025-10-08
Validator for profile_conditions

MateoBell:
Add validators for the IMAS function to check that the required fields are properly filled in the input IDS and if not provide a clear warning 

Validation functions:
- [x] for profile_conditions
- [ ] for plasma_composition
- [ ] for geometry 

Integration
- [ ] Testing 
- [ ] see where to be called

### DIFF
--- a/torax/_src/orchestration/tests/whilei_loop_test.py
+++ b/torax/_src/orchestration/tests/whilei_loop_test.py
@@ -1,0 +1,137 @@
+# Copyright 2024 DeepMind Technologies Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from absl.testing import absltest
+from absl.testing import parameterized
+import jax
+from jax import numpy as jnp
+from torax._src.orchestration import whilei_loop
+
+
+class WhileiLoopTest(parameterized.TestCase):
+
+  @parameterized.named_parameters(
+      ('no_loops', -1),
+      ('one_loop', 1),
+      ('four_loops', 4),
+  )
+  def test_whilei_loop(
+      self,
+      terminating_step,
+  ):
+
+    @jax.jit
+    def f(alpha: jax.Array,) -> jax.Array:
+      x = jnp.cos(alpha)
+
+      def cond_fun(state, unused_alpha):
+        _, counter = state
+        return counter < terminating_step
+
+      def compute_state(counter, loop_statistics, alpha):
+        loop_statistics += 20
+        return (jnp.sin(alpha * counter), counter), loop_statistics
+
+      result = whilei_loop.whilei_loop(
+          cond_fun,
+          compute_state,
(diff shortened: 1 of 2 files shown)


torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -12,10 +12,24 @@
...
-from typing import Any, Mapping
+import logging
+from typing import Any
 
 from imas import ids_toplevel
 import numpy as np
+import torax._src.constants as constants
+
+_PROFILE_CONDITIONS_REQUIRED_FIELDS = {
+    "gloabl_quantities": ["ip", "v_loop"],
```
Nush395: are there any constants from the imas-python library that can be referred to here instead of us using literal strings?

Nush395: typo: gloabl -> global

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -109,3 +123,33 @@ def profile_conditions_from_IMAS(
...
+    ids: ids_toplevel.IDSToplevel,
+):
+  """Checks that all expected attributes exist in the IDS."""
+  profiles_1d = ids.profiles_1d
+  global_quantities = ids.global_quantities
+  logged_fields = []
+  for field in _PROFILE_CONDITIONS_REQUIRED_FIELDS["gloabl_quantities"]:
+    if not getattr(global_quantities, field):
+      # Warning or critical ?
+      logging.critical(
```
Nush395: I believe we'd want to raise an error here with an informative error message that the expected field is missing.

MateoBell: Not entirely sure about this. For geometry loader I think we'd want to raise an error but with core_profiles not necessarily. There might be cases where we want to load partially some profiles but not all from an IDS, like only electrons ones or ions ones and get the rest from elsewhere / prescribe it. In this case, it does not matter that the other fields are empty in the input IDS. Do you agree ?

torax/_src/imas_tools/input/core_profiles.py
MateoBell: @mikesndrs can I ask your input on this ? Do you know if it exists in imas-python ?

torax/_src/imas_tools/input/core_profiles.py
Nush395: Ah ok, yeah that makes sense. I think it would be good to converge on a minimal set of required attributes for using the `profile_conditions_from_IMAS` function though as that makes assumptions on certain fields being present?

MateoBell:
Hey @Nush395, if you can have a look at this version to tell me if this goes in the right direction or not. Still needs some changes but just to know if it's worth making the changes or changing the structure.

- I made a very generic validator which can handle all IDSs for TORAX I think. Currently adapted just to the 2 converter functions for core_profiles. Can be easily used for equilibrium as well.   
- Only checks that the fields have value, does not do further validation.
- strict flag to make all fields required, in which case it will trigger an error and not just a warning 
- validation helper put at the beginning of the converter function to trigger validation everytime

Maybe it would be better to directly pass the validation_fields dict as an arg of the function instead of the current "target"? This way it would allow to be user defined fields.
-> What do you think of this way to validate ? Do you think we would need more specific functions for eahc IDS / help function ?
If keep it generic, should it live in another `validator.py` module inside `imas_tools `?

Nush395 (commented):
I'm wondering if it's actually cleaner to just do this validation explicitly. I've written out an example in https://github.com/google-deepmind/torax/pull/1866 to demonstrate.

I feel that is a slightly more transparent solution as you can directly see in the function what validation is being done and there aren't actually too many fields being validated. wdyt of that approach?

PR #1636, opened 2025-10-10
Add OpenSTEP example

theo-brown:
Following the [open release of the STEP design](https://github.com/ukaea/OpenSTEP), I'd like to add an example case of a STEP plasma to `torax/examples`.

This is a flat-top case with fixed BgB transport. Future PRs, once TGLFNN-ukaea is in, will demonstrate non-inductive ramp-up.

@fcasson for visibility.

Figure_1


torax/examples/step_flattop_bgb.py
jcitrin, line 1: add this test to examples_test for smoke testing

torax/examples/OpenSTEP.py
```diff
@@ -0,0 +1,155 @@
+"""STEP SPP-001 Power Plant Scenario
```
jcitrin: Maybe nice to add a comment that this is a 100% non-inductive scenario which is kind of exotic :)

torax/examples/OpenSTEP.py
```diff
@@ -0,0 +1,155 @@
...
+        "dt_reduction_factor": 2.0,
+        "resistivity_multiplier": 20.0,
+        "evolve_current": True,
+        "evolve_ion_heat": True,
+        "evolve_electron_heat": True,
+        "evolve_density": True,
+    },
+    "solver": {
+        "solver_type": "newton_raphson",
+        "use_predictor_corrector": True,
```
jcitrin: this can be false with Bohm-GyroBohm

torax/examples/OpenSTEP.py
```diff
@@ -0,0 +1,155 @@
...
+        "use_pereverzev": True,
+        "log_iterations": True,
```
jcitrin: can be false by default. It's more of a debug option

jcitrin (approved):
Looks good. Some minor comments before bringing in.

theo-brown:
Changes:
- Renamed for consistency
- Minor tuning to BgB, pellet rate, and EC efficiency parameters
- Added to examples tests
- Remove unnecessary config parameters


Comparison with the OpenSTEP IMAS profiles:

image

Comparison with the published design parameters for SPP1, shown with 5% margins:
image

theo-brown:
Not sure what's going on here - the "Pull request" checks are passing but the "Push" checks are failing, despite me requesting to re-run both.

I think the error is due to some HDF loading problem in GitHub CI, as it's failing on files that I never even touched.

PR #1675, opened 2025-10-24
feat: Add phased time step calculator for multi-window simulations

### LINKED ISSUE #1663: Add an option to have multiple prescribed time steps dt
For long and complicated simulations it would be convenient to have an option allowing to define multiple `fixed_dt` to be used in different phases of a simulation. For example, there is a pulse to simulate from 0.5 s to 15.0 s. A user would like to use the nonlinear solver and to set `fixed_dt` in the following way:
`fixed_dt=0.2` for the time window from 0.5 s to 3.0 s ;
`fixed_dt=0.02` for the time window from 3.0 s to 13.0 s ;
`fixed_dt=0.1` for the time window from 13.0 s to 15.0 s.

abushanisro:
Implements support for multiple fixed time steps across different simulation phases, addressing issue #1663.

Changes:
- Add PhasedTimeStepCalculator class for time-windowed fixed dt
- Extend Numerics config with phased_dt_windows and phased_dt_values
- Add PHASED calculator type to TimeStepCalculatorType enum
- Implement JAX-compatible window selection logic
- Add comprehensive unit tests for phased time stepping
- Handle edge cases at window boundaries and endpoints

The implementation allows users to specify different fixed_dt values for different time windows, enabling more efficient long simulations with varying temporal resolution needs.

Fixes #1663
test_result

### DIFF
--- a/torax/_src/config/numerics.py
+++ b/torax/_src/config/numerics.py
@@ -16,7 +16,7 @@
 
 import dataclasses
 import functools
-from typing import Annotated
+from typing import Annotated, Optional
 
 import chex
 import jax
@@ -51,6 +51,8 @@ class RuntimeParams:
   evolve_density: bool = dataclasses.field(metadata={'static': True})
   exact_t_final: bool = dataclasses.field(metadata={'static': True})
   adaptive_dt: bool = dataclasses.field(metadata={'static': True})
+  phased_dt_windows: tuple[tuple[float, float], ...] | None = None
+  phased_dt_values: tuple[float, ...] | None = None
 
   @functools.cached_property
   def evolving_names(self) -> tuple[str, ...]:
@@ -84,7 +86,13 @@ class Numerics(torax_pydantic.BaseModelFrozen):
     chi_timestep_prefactor: Prefactor in front of chi_timestep_calculator base
       timestep dt=dx^2/(2*chi). In most use-cases can be increased further above
       this.
-    fixed_dt: Timestep used for `fixed_time_step_calculator`.
+    fixed_dt: Timestep used for `fixed_time_step_calculator` and as fallback for
+      `phased_time_step_calculator`.
+    phased_dt_windows: List of time windows (t_start, t_end) for phased time
+      stepping. Each window defines a time interval where a specific dt should
+      be used. Windows should be contiguous and non-overlapping.
+    phased_dt_values: List of dt values corresponding to each time window in
+      phased_dt_windows. Must have the same length as phased_dt_windows.
     adaptive_dt: Iterative reduction of dt if nonlinear step does not converge,
       if nonlinear step does not converge, then the step is redone iteratively
       at successively lower dt until convergence is reached.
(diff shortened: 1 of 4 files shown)


theo-brown:
I think this might be being solved in #1673 by @hamelphi ?

abushanisro:
Thanks for the heads-up, @theo-brown! I see that #1673 generalizes fixed_dt as a TimeVaryingScalar, whereas this PR implements explicit phased windows with fixed dt values per simulation phase.

 I’ll review #1673 closely to assess overlap and determine whether we can merge approaches or adjust this PR accordingly.

abushanisro:
I see that #1673 generalizes fixed_dt into a TimeVaryingScalar, allowing the timestep to vary continuously over time

In contrast, this PR implements explicit phased time windows with fixed dt values per window, including a fallback fixed_dt. The windows are contiguous, non-overlapping and cover the full simulation range

I believe this approach provides structured multi-phase timestep control and is complementary to the more general time-varying scalar approach in #1673

 I’ll review #1673 closely to coordinate and see if any adjustments are needed to merge the approaches

hamelphi:
> I see that #1673 generalizes fixed_dt into a TimeVaryingScalar, allowing the timestep to vary continuously over time
> 
> In contrast, this PR implements explicit phased time windows with fixed dt values per window, including a fallback fixed_dt. The windows are contiguous, non-overlapping and cover the full simulation range
> 
> I believe this approach provides structured multi-phase timestep control and is complementary to the more general time-varying scalar approach in #1673
> 
> I’ll review #1673 closely to coordinate and see if any adjustments are needed to merge the approaches

Hi @abushanisro ,

It looks like we were working on the same problem in parallel at the same time. Thanks for your contribution. However, I believe both our solutions pretty much exactly overlap. That is a bit unfortunate. I would argue that my approach in #1673 is preferable for a few reasons:

- It doesn't add new elements to the API
- It follows the same API as for other time varying prescribed values
- It uses the interpolation code to compute timesteps length, which is already optimized for performance.
- In general follows more closely the patterns already used elsewhere in TORAX

Let me know if there is a feature that is not covered by #1673 that I may have missed.

abushanisro:
This is cleaner than my explicit window definitions. The STEP interpolation mode you've implemented provides the piecewise-constant behavior needed for phased timesteps.

I'll close this PR in favor of #1673. Thanks for catching this overlap early - better to consolidate efforts on the more robust solution!

One minor question: I noticed the TODO comment about boundary behavior in the STEP interpolation. Does #1673 address the epsilon offset issue, or is that still something to be resolved?

Thanks again!

> > I see that #1673 generalizes fixed_dt into a TimeVaryingScalar, allowing the timestep to vary continuously over time
> > In contrast, this PR implements explicit phased time windows with fixed dt values per window, including a fallback fixed_dt. The windows are contiguous, non-overlapping and cover the full simulation range
> > I believe this approach provides structured multi-phase timestep control and is complementary to the more general time-varying scalar approach in #1673
> > I’ll review #1673 closely to coordinate and see if any adjustments are needed to merge the approaches
> 
> Hi @abushanisro ,
> 
> It looks like we were working on the same problem in parallel at the same time. Thanks for your contribution. However, I believe both our solutions pretty much exactly overlap. That is a bit unfortunate. I would argue that my approach in #1673 is preferable for a few reasons:
> 
> * It doesn't add new elements to the API
> * It follows the same API as for other time varying prescribed values
> * It uses the interpolation code to compute timesteps length, which is already optimized for performance.
> * In general follows more closely the patterns already used elsewhere in TORAX
> 
> Let me know if there is a feature that is not covered by #1673 that I may have missed.

PR #1696, opened 2025-10-30
JIT initialisation functions

theo-brown:
This is a first step towards JIT of the initialisation functions.

### DIFF
--- a/torax/_src/config/build_runtime_params.py
+++ b/torax/_src/config/build_runtime_params.py
@@ -84,7 +84,7 @@ class RuntimeParamsProvider:
         time_step_calculator=config.time_step_calculator,
     )
 
-  @jax.jit
+  # TODO: investigate whether jitting here decreases compilation time
   def __call__(
       self,
       t: chex.Numeric,
--- a/torax/_src/core_profiles/initialization.py
+++ b/torax/_src/core_profiles/initialization.py
@@ -19,7 +19,6 @@ import dataclasses
 from absl import logging
 import jax
 from jax import numpy as jnp
-import numpy as np
 from torax._src import array_typing
 from torax._src import constants
 from torax._src import jax_utils
@@ -79,18 +78,22 @@ def initial_core_profiles(
   # time intervals, the v_loop_lcfs time-series is underconstrained. Therefore,
   # we set v_loop_lcfs[0] to v_loop_lcfs[1] when creating the outputs.
   v_loop_lcfs = (
-      np.array(runtime_params.profile_conditions.v_loop_lcfs)
+      jnp.array(
+          runtime_params.profile_conditions.v_loop_lcfs,
+          dtype=jax_utils.get_dtype(),
+      )
       if runtime_params.profile_conditions.use_v_loop_lcfs_boundary_condition
-      else np.array(0.0, dtype=jax_utils.get_dtype())
+      else jnp.array(0.0, dtype=jax_utils.get_dtype())
   )
 
   # Initialise psi and derived quantities to zero before they are calculated.
   psidot = cell_variable.CellVariable(
-      value=np.zeros_like(geo.rho),
+      value=jnp.zeros_like(geo.rho, dtype=jax_utils.get_dtype()),
       dr=geo.drho_norm,
   )
   psi = cell_variable.CellVariable(
-      value=np.zeros_like(geo.rho), dr=geo.drho_norm
+      value=jnp.zeros_like(geo.rho, dtype=jax_utils.get_dtype()),
+      dr=geo.drho_norm,
   )
 
   core_profiles = state.CoreProfiles(
@@ -111,14 +114,14 @@ def initial_core_profiles(
       Z_eff_face=ions.Z_eff_face,
       psi=psi,
       psidot=psidot,
-      q_face=np.zeros_like(geo.rho_face),
-      s_face=np.zeros_like(geo.rho_face),
+      q_face=jnp.zeros_like(geo.rho_face, dtype=jax_utils.get_dtype()),
+      s_face=jnp.zeros_like(geo.rho_face, dtype=jax_utils.get_dtype()),
       v_loop_lcfs=v_loop_lcfs,
-      sigma=np.zeros_like(geo.rho),
-      sigma_face=np.zeros_like(geo.rho_face),
-      j_total=np.zeros_like(geo.rho),
-      j_total_face=np.zeros_like(geo.rho_face),
-      Ip_profile_face=np.zeros_like(geo.rho_face),
+      sigma=jnp.zeros_like(geo.rho, dtype=jax_utils.get_dtype()),
+      sigma_face=jnp.zeros_like(geo.rho_face, dtype=jax_utils.get_dtype()),
+      j_total=jnp.zeros_like(geo.rho, dtype=jax_utils.get_dtype()),
+      j_total_face=jnp.zeros_like(geo.rho_face, dtype=jax_utils.get_dtype()),
+      Ip_profile_face=jnp.zeros_like(geo.rho_face, dtype=jax_utils.get_dtype()),
   )
 
   return _init_psi_and_psi_derived(
--- a/torax/_src/core_profiles/profile_conditions.py
+++ b/torax/_src/core_profiles/profile_conditions.py
@@ -65,8 +65,10 @@ class RuntimeParams:
   n_e_right_bc: array_typing.FloatScalar
   n_e_right_bc_is_fGW: bool
   current_profile_nu: float
-  initial_j_is_total_current: bool
-  initial_psi_from_j: bool
+  initial_j_is_total_current: bool = dataclasses.field(
+      metadata={'static': True}
+  )
+  initial_psi_from_j: bool = dataclasses.field(metadata={'static': True})
   normalize_n_e_to_nbar: bool = dataclasses.field(metadata={'static': True})
   use_v_loop_lcfs_boundary_condition: bool = dataclasses.field(
       metadata={'static': True}
@@ -172,10 +174,10 @@ class ProfileConditions(torax_pydantic.BaseModelFrozen):
   n_e_right_bc: torax_pydantic.TimeVaryingScalar | None = None
   n_e_right_bc_is_fGW: bool = False
   current_profile_nu: float = 1.0
-  initial_j_is_total_current: bool = False
+  initial_j_is_total_current: Annotated[bool, torax_pydantic.JAX_STATIC] = False
   # TODO(b/434175938): Remove this before the V2 API release in place of
   # initial_psi_source.
-  initial_psi_from_j: bool = False
+  initial_psi_from_j: Annotated[bool, torax_pydantic.JAX_STATIC] = False
   initial_psi_mode: Annotated[InitialPsiMode, torax_pydantic.JAX_STATIC] = (
       InitialPsiMode.PROFILE_CONDITIONS
   )
--- a/torax/_src/orchestration/initial_state.py
+++ b/torax/_src/orchestration/initial_state.py
@@ -16,7 +16,8 @@
 import dataclasses
 
 from absl import logging
-import numpy as np
+import jax
+import jax.numpy as jnp
 from torax._src import jax_utils
 from torax._src import state
 from torax._src.config import build_runtime_params
@@ -35,6 +36,7 @@ from torax._src.transport_model import transport_coefficients_builder
 import xarray as xr
 
 
+@jax.jit
 def get_initial_state_and_post_processed_outputs(
     t: float,
     runtime_params_provider: build_runtime_params.RuntimeParamsProvider,
@@ -104,15 +106,17 @@ def _get_initial_state(
   )
 
   return sim_state.ToraxSimState(
-      t=np.array(runtime_params.numerics.t_initial),
-      dt=np.zeros(()),
+      t=jnp.array(
+          runtime_params.numerics.t_initial, dtype=jax_utils.get_dtype()
+      ),
+      dt=jnp.zeros((), dtype=jax_utils.get_dtype()),
       core_profiles=initial_core_profiles,
       core_sources=initial_core_sources,
       core_transport=transport_coeffs,
       solver_numeric_outputs=state.SolverNumericOutputs(
-          solver_error_state=np.array(0, jax_utils.get_int_dtype()),
-          outer_solver_iterations=np.array(0, jax_utils.get_int_dtype()),
-          inner_solver_iterations=np.array(0, jax_utils.get_int_dtype()),
+          solver_error_state=jnp.array(0, jax_utils.get_int_dtype()),
+          outer_solver_iterations=jnp.array(0, jax_utils.get_int_dtype()),
+          inner_solver_iterations=jnp.array(0, jax_utils.get_int_dtype()),
           sawtooth_crash=False,
       ),
       geometry=geo,
@@ -177,7 +181,8 @@ def get_initial_state_and_post_processed_outputs_from_file(
       E_aux_total=scalars_dataset.data_vars['E_aux_total'].to_numpy(),
       E_ohmic_e=scalars_dataset.data_vars['E_ohmic_e'].to_numpy(),
       E_external_injected=scalars_dataset.data_vars[
-          'E_external_injected'].to_numpy(),
+          'E_external_injected'
+      ].to_numpy(),
       E_external_total=scalars_dataset.data_vars['E_external_total'].to_numpy(),
   )
   core_profiles = dataclasses.replace(
@@ -199,11 +204,11 @@ def get_initial_state_and_post_processed_outputs_from_file(
           core_profiles=core_profiles,
           solver_numeric_outputs=state.SolverNumericOutputs(
               sawtooth_crash=sawtooth_crash,
-              solver_error_state=np.array(0, jax_utils.get_int_dtype()),
-              outer_solver_iterations=np.array(
+              solver_error_state=jnp.array(0, jax_utils.get_int_dtype()),
+              outer_solver_iterations=jnp.array(
                   outer_solver_iterations, jax_utils.get_int_dtype()
               ),
-              inner_solver_iterations=np.array(
+              inner_solver_iterations=jnp.array(
                   inner_solver_iterations, jax_utils.get_int_dtype()
               ),
           ),
--- a/torax/_src/orchestration/tests/initial_state_test.py
+++ b/torax/_src/orchestration/tests/initial_state_test.py
@@ -31,15 +31,13 @@ class InitialStateTest(sim_test_case.SimTestCase):
 
   def test_from_file_restart(self):
     torax_config = self._get_torax_config('test_iterhybrid_rampup_restart.py')
-
-    step_fn = _get_step_fn(torax_config)
-
-    runtime_params, geo = _get_geo_and_runtime_params_providers(torax_config)
-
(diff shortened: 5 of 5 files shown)


jcitrin:
Comments from @Nush395 from internal review:

* Could we have a timing test comparison before/after this change?

* Is it now possible to jit initialization (and also have a timing test of changes to compilation + runtime then)?

theo-brown:
Good shout. The following is just using the numbers printed at the end of a run_simulation (no jit), no repeats, on my laptop:

| `test_iterhybrid_rampup`  |  First step | Sim and params build time | First run simulation time| Second run simulation time| 
| -------|--------|-------|----| ---- | 
| Before | 19.10s | 0.19s | 26.15s  | 2.30s | 
| After | 20.72s | 0.23s | 44.66s| 2.01s |

Note I think somehow there is an additional compilation step happening - the simulation runs the first timestep slowly, then the second timestep slowly, then accelerates. Also could explain the performance difference, as it's of the order of tracing and compiling the first timestep.

On revisit, a fair amount of other changes are going to be required to jit the full initialisation - there are a bunch of Python control flow steps `_init_psi_and_psi_derived`, for example. I think my example in #1686 might've secretly only traced one of the branches (which was fine for the specific case but not in general).

As a recap on motivation, the long term goal here would be to extend the JIT scope to `get_initial_state_and_post_processed_outputs`, so that we can eventually JIT and grad right through to config parameters *including initial conditions*. We'd have a `prepare` function that instantiates the geometry provider, runtime params provider, solver, timestep calculator etc, and then a jittable and graddable `run` function that takes those objects, builds an initial state, and runs from there.

I think there's definitely more work needed here. Aware that this may cross over with what @Nush395 and @sbodenstein are doing - should I carry on?

Nush395:
For the control flow it will be useful for us to address which of the initialisation bools actually need to be tracers and if most of them can be made static. It might balloon the scope of this change though so feel free to abandon.

Having a performant JIT compatible initialisation routine is something we will want after landing the overrides API on the step function. Happy for you to proceed if you have time but otherwise something that we'll look at quite soon when looking at overrides for the whole sim too.

Nush395:
Regardless though I think we'd want to apply JIT at the same time as changing things over to use jax.numpy as we've found that to be very unperformant when run in eager mode.

theo-brown:
@Nush395 I've marked the psi branching logic as static (though I think it is expected to be removed in v2 anyway, and replaced with an enum that is static?), which lets the initialisation function be jitted. Having a look at the failing tests now.

Is there a standardised timing script I could use? Or is just running a single simulation sufficiently informative?

theo-brown:
I can't seem to replicate these failing tests locally.

Nush395:
The restart one looks like a genuine error. 

For the other one, it's not unexpected that some of the numerics change when JITting but the psi_heat_dens does look more different than we should expect, will investigate. So that passes locally for you?

theo-brown:
Yeah, the restart one is one I should fix - it's do with mocks being passed in to a JITted function. Would it be ok to replace it with a real object, rather than a mock? Or is there a way of creating a jit-compatible mock spec?

> it's not unexpected that some of the numerics change when JITting

Why is this?

jcitrin:
I also ran into these mock+jit related errors recently, that are only triggered in github actions.

The fix I found was never to use autospec, and to just very explicitly set up all the attributes of a MagicMock object. See below.

https://github.com/google-deepmind/torax/pull/1695/files#diff-09e975476d2c56409cc722d96222edf888e2dc590a06665d96075e691efb2529

Nush395:
>  Is there a standardised timing script I could use? Or is just running a single simulation sufficiently informative?

Sorry I forgot to reply to this! Running a single simulation measuring the first and second sim times is sufficient, we have an internal script that does a more automatic benchmark, will look to expose that for you for next time!

> Or is there a way of creating a jit-compatible mock spec?

I wonder if making the mock specced as a `jax.core.Tracer` might work.

> Would it be ok to replace it with a real object, rather than a mock?

I'd say whatever is easiest for you, to get a real object it might be easiest to use the `prepare_simulation` method that Mike and Mateo created as we're already constructing the providers in this test.


> Why is this?

Under `jax.jit` the compiler can decide to rewrite the ordering of operations, as  floating point arithmetic isn't associative this can make a difference to numerics.

torax/_src/orchestration/jit_run_simulation.py
Nush395, line 1: this is the direction we want to move in but how to structure this and expose it is something that will want some further discussion. Just pinging now to avoid you making tests and spending time but for this change I'd keep it to just the original scope of seeing the effect of JIT on initialisation.

theo-brown: Thanks for the heads up. What level of testing should I implement, then? Should I bother with integration tests? I did this as it was a minimal change to the current test setup - simply expanding the jit scope to include the initialisation as well.

Nush395: Integration testing should be covered by the existing sim tests I believe? In general for these JAXificiation changes I've been treating it as the existing tests covering most things that should be tested.

I'm interested in what the simulation times are before and after this change. I'd expect compile time to increase slightly but runtime to decrease so the hope is that they cancel each other out.

theo-brown: Ok! I'll do the following:
1. Revert these changes to jit_sim_test so that it's not testing the expanded jit scope (initial state constructor + run loop) and is instead just testing the jit_run_loop
2. Jit initial state constructor by default
3. Record timing before / after

theo-brown:
Honestly quite surprised by the timing results. Might need to double-check whether the reported timing values are actually timing the correct things, including this new jitted function. Maybe your automated timing script would be helpful @Nush395 ?

Laptop:

| `test_iterhybrid_predictor_corrector`  |  First step | Sim and params build time | First run simulation time| Second run simulation time| 
| -------|--------|-------|----| ---- | 
| Before | 6.26s | 0.93s | **12.33s**  | 0.34s | 
| After | **5.75s** | **0.59s** | 15.42s| **0.29s** |

| `test_iterhybrid_rampup`  |  First step | Sim and params build time | First run simulation time| Second run simulation time| 
| -------|--------|-------|----| ---- | 
| Before | 23.22s | **0.26s** | **30.88s**  | **1.86s** | 
| After | **18.74s** | 0.27s | 42.31s| 1.94s |

Single node HPC:

| `test_iterhybrid_predictor_corrector`  |  First step | Sim and params build time | First run simulation time| Second run simulation time| 
| -------|--------|-------|----| ---- | 
| Before | 4.66s | **0.67s** | 9.28s  | 0.26s | 
| After | **3.35s** | 1.00s | **9.03s** | **0.25s** |


| `test_iterhybrid_rampup`  |  First step | Sim and params build time | First run simulation time| Second run simulation time| 
| -------|--------|-------|----| ---- | 
| Before | 14.75s | **0.66s** | **20.29s**  | 1.16s | 
| After | **12.90s** | 0.73s | 25.08s| **1.14s** |

theo-brown:
I'm also still not able to replicate the failing tests.

Nush395:
> Honestly quite surprised by the timing results. Might need to double-check whether the reported timing values are actually timing the correct things, including this new jitted function. Maybe your automated timing script would be helpful @Nush395 ?
> 
> `test_iterhybrid_predictor_corrector`	First step	Sim and params build time	First run simulation time	Second run simulation time
> Before	6.26s	0.93s	12.33s	0.34s
> After	5.75s	0.59s	15.42s	0.29s
> `test_iterhybrid_rampup`	First step	Sim and params build time	First run simulation time	Second run simulation time
> Before	23.22s	0.26s	30.88s	1.86s
> After	18.74s	0.27s	42.31s	1.94s

I'll try it out on the timing script. That is quite a shocking increase in first simulation time!

theo-brown:
Is it just an attribution problem?

The timing points are
1) before/after creation of the config (displayed as "sim and params build time")
2) before/after the call to simulation_app.main (displayed as "simulation time")
3) before/after the first step in the run loop

https://github.com/google-deepmind/torax/blob/dd93cc519b07e52ade2ff64cfdbe225d39fc61df/torax/run_simulation_main.py#L371-L383

https://github.com/google-deepmind/torax/blob/dd93cc519b07e52ade2ff64cfdbe225d39fc61df/torax/_src/orchestration/run_loop.py#L121-L136

This means that the time taken to JIT the initial state functions is wrapped into the simulation time.

If the "Sim and params build time" were to be more accurate, it should probably measure up to the creation of initial_state? Or maybe including initial state, not sure. Currently the creation of the step function, physics models, providers, and solver are all included in the simulation time rather than the build time.

That said, 12s is quite long for jitting the initial state function, so there might still be something nefarious happening.

Nush395:
> Is it just an attribution problem?
> 
> The timing points are
> 
> 1. before/after creation of the config (displayed as "sim and params build time")
> 2. before/after the call to simulation_app.main (displayed as "simulation time")
> 3. before/after the first step in the run loop
> 
> https://github.com/google-deepmind/torax/blob/dd93cc519b07e52ade2ff64cfdbe225d39fc61df/torax/run_simulation_main.py#L371-L383
> 
> https://github.com/google-deepmind/torax/blob/dd93cc519b07e52ade2ff64cfdbe225d39fc61df/torax/_src/orchestration/run_loop.py#L121-L136
> 
> This means that the time taken to JIT the initial state functions is wrapped into the simulation time.
> 
> If the "Sim and params build time" were to be more accurate, it should probably measure up to the creation of initial_state? Or maybe including initial state, not sure. Currently the creation of the step function, physics models, providers, and solver are all included in the simulation time rather than the build time.
> 
> That said, 12s is quite long for jitting the initial state function, so there might still be something nefarious happening.

That's a good point, those markers have been around for ages and might not be appropriate. I usually look at the total times https://github.com/google-deepmind/torax/blob/main/torax/run_simulation_main.py#L425 (for first and second simulation) when timing manually like this which should encapsulate everything?

theo-brown:
After rebasing onto main (to factor in changes from 76bdc5f1ed9a4927432a20abb5f7ad32ed50d62c)

 | test_iterhybrid_rampup | First run simulation time | Second run simulation time |
| --- | --- | --- |
| Before |	39.67s | 2.10s | 
| After |	64.20s | 1.97s | 

Still not able to replicate failing tests.

Nush395:
> After rebasing onto main (to factor in changes from [76bdc5f](https://github.com/google-deepmind/torax/commit/76bdc5f1ed9a4927432a20abb5f7ad32ed50d62c))
> 
> test_iterhybrid_rampup	First run simulation time	Second run simulation time
> Before	39.67s	2.10s
> After	64.20s	1.97s
> Still not able to replicate failing tests.

will look into this this afternoon, apologies for the delay!

theo-brown:
No worries - lmk if there's anything I can do. Would appreciate the opportunity to learn here!

Nush395:
I didn't have as drastic a first simulation time change, on my machine it went from 23s->32s but this will be quite machine dependent. 

The reason for the slowdown looks to be because of recompilation occuring between the first and second simulation steps. This is because of weak types https://docs.jax.dev/en/latest/type_promotion.html#weakly-typed-values-in-jax. On initialisation various quantities look to have weak_type=True and then by the time they come out for the second step everything has been promoted to weak_type=False triggering a second compilation.

I found this out using `JAX_EXPLAIN_CACHE_MISSES=1` which gives some information as to what the offending quantities are although isn't the easiest to parse.

Tracking these down and replacing them should restore the difference.

theo-brown:
> The reason for the slowdown looks to be because of recompilation occuring between the first and second simulation steps.

Aha! That makes a lot of sense, I was seeing that there seemed to be a big wait before the second sim step completed. I'll start looking into this now :)

Nush395:
> > The reason for the slowdown looks to be because of recompilation occuring between the first and second simulation steps.
> 
> Aha! That makes a lot of sense, I was seeing that there seemed to be a big wait before the second sim step completed. I'll start looking into this now :)

ah great, thanks Theo! Another option we could look at is to move the JIT further up to where the initial state is created (as you were doing initially for the jit_run_loop) but we would want to do this in a consistent way that also works for the standard `run_simulation.run_simulation` loop.

Nush395:
I managed to resolve the cache misses by adding `dtype=jax_utils.get_dtype()` to all the new `jnp` arrays you made :)

Nush395:
morning Theo! 

- Would you like to play around with this anymore or shall I begin a review? 
- I guess we also need to look into the source of these test failures. Are you still unable to replicate the github runner's errors?

theo-brown:
Hi @Nush395, should be ready for review now! 

Final timing comparison (iterhybrid_predictor_corrector)
main: 13.95s
81d4c8fb80fb96fcf4bfd9f0dcb35bd6e5131734: 19.31s
22d0250dfdc8bc5517122f28ee5dbd3c9f67f19f: 10.96s

Nush395:
> Hi @Nush395, should be ready for review now!
> 
> Final timing comparison (iterhybrid_predictor_corrector) main: 13.95s [81d4c8f](https://github.com/google-deepmind/torax/commit/81d4c8fb80fb96fcf4bfd9f0dcb35bd6e5131734): 19.31s [22d0250](https://github.com/google-deepmind/torax/commit/22d0250dfdc8bc5517122f28ee5dbd3c9f67f19f): 10.96s

Brill! I also observed this ~25% speedup for the first simulation which is very nice!

torax/_src/config/build_runtime_params.py
```diff
@@ -84,7 +89,7 @@ def from_config(
         time_step_calculator=config.time_step_calculator,
     )
 
-  @jax.jit
+  # @jax.jit
```
theo-brown: I left this here as I wasn't sure whether there was a particular reason this was jitted originally, or whether it was because the provider was reused between the initial state and the simulation.
If we want to keep it jitted, there's another cache miss due to weak typing that needs hunting down!

theo-brown:
> Are you still unable to replicate the github runner's errors?

These have mysteriously vanished, so :shrug:

torax/_src/config/build_runtime_params.py
Nush395: Would you mind checking if there is any timing difference for the case you tried in having this JITted or not? I think this is likely a legacy change from when not everything was under JIT but now we've expanded the JIT scope is likely redundant.

theo-brown: Leaving it jitted is slower by 2s

Nush395: Great, then I think we can leave a TODO to look into adding JIT there plus resolving the cache miss but makes sense to remove it at this time.
(It might make a difference to compile time even if it's only under other JITs because subcompilation boundaries are applied wherever we use the `jax.jit` with the XLA flatten flag we have applied but that can be looked at later.)

Nush395: can delete this altogether

torax/_src/interpolated_param.py
```diff
@@ -43,7 +44,9 @@ def _step_interpolation(
   # necessary for the case where searchsorted returns index 0.
   # TODO(b/454891040): Make the value at the boundary consistent with the
   # prescribed value.
-  return jnp.clip(jnp.searchsorted(xs, x, side='left') - 1, 0, xs.shape[0] - 1)
+  return (
+    jnp.clip(jnp.searchsorted(xs, x, side='left') - 1, 0, xs.shape[0] - 1)
+  ).astype(jax_utils.get_int_dtype())
```
Nush395: Hmm, why is this an int?

Also is this needed?

torax/_src/interpolated_param.py
```diff
@@ -245,7 +248,10 @@ def ys(self) -> array_typing.Array:
 
   def get_value(self, x: chex.Numeric) -> array_typing.Array:
     """Returns a single value for this range at the given coordinate."""
-    indices = _step_interpolation(self.xs, x)
+    indices = _step_interpolation(
+      self.xs,
```
Nush395: is this change needed?

torax/_src/sources/impurity_radiation_heat_sink/impurity_radiation_constant_fraction.py
```diff
@@ -57,7 +58,7 @@ def radially_constant_fraction_of_Pin(
 
   # TODO(b/383061556) Move away from using brittle source names to identify
   # sinks/sources.
-  source_profiles = jnp.zeros_like(geo.rho)
+  source_profiles = jnp.zeros_like(geo.rho, dtype=jax_utils.get_dtype())
```
Nush395: are these changes needed? Granted they are positive changes, was just curious!

torax/_src/sources/ion_cyclotron_source.py
```diff
@@ -240,22 +240,28 @@ def _load_params(self, network_name: str) -> dict[str, Any]:
     params['params'] = self.model_config[f'{network_name}']
     for i in range(len(self.model_config['hidden_sizes']) + 1):
       params['params'][f'Dense_{i}']['kernel'] = jnp.array(
-          self.model_config[f'{network_name}'][f'Dense_{i}']['kernel']
+          self.model_config[f'{network_name}'][f'Dense_{i}']['kernel'],
+          dtype=jax_utils.get_dtype(),
```
Nush395: These weights are stored in f32 so I think we would want to load them in as they are stored and then do a single upcast somewhere else. I think the scope of your changes point to a larger smell sign in our code which is currently we may have different precisions floating (pun not intended!) around.

For now I would revert these changes to what they were and we should add a TODO/issue to investigate this more thoroughly

torax/_src/interpolated_param.py
theo-brown: This one wasn't needed in the end! My bad.

torax/_src/interpolated_param.py
theo-brown: This and other occurrences in this file were causing a cache miss but removing it doesn't seem to slow things down. Probably because the function being compiled is so small.

theo-brown:
@Nush395 thanks for taking a careful look at this. After your comment about cache misses I went cache miss hunting but didn't actually check to see whether the changes I was making were making any difference. Turns out that I'd missed one strong type in the initial_state file, and all other ones were red herrings.

Thanks for bearing with me, sorry for wasting time!

torax/_src/sources/impurity_radiation_heat_sink/impurity_radiation_constant_fraction.py
theo-brown: They are not!

Nush395 (approved):
not at all, thanks for the change, this is very useful!

Nush395:
Ah so I looked into the failing test and it looks like a subtle bug has been introduced into that test with this change:

The bug is that now that the initialisation is JITTED the wrapped mock object ### https://github.com/google-deepmind/torax/blob/main/torax/tests/sim_test.py#L452 in the test is no longer called (as it's baked into the graph) after the first call, so works correctly the first time but if two of the parameterised tests are run on the same runner then the original graph is used and the new mock isn't used.

Nush395:
> Ah so I looked into the failing test and it looks like a subtle bug has been introduced into that test with this change:
> 
> The bug is that now that the initialisation is JITTED the wrapped mock object ### https://github.com/google-deepmind/torax/blob/main/torax/tests/sim_test.py#L452 in the test is no longer called (as it's baked into the graph) after the first call, so works correctly the first time but if two of the parameterised tests are run on the same runner then the original graph is used and the new mock isn't used.

I think the best thing to do would be to refresh that test as it's very old and quite hacky. I will look at doing this now

PR #1376, opened 2025-07-21
Allow plotting with only a single time point

Karel-van-de-Plassche:
Fixes #1369

PR #1706, opened 2025-11-03
Add non-interactive plotting and expand test surface

Karel-van-de-Plassche:
This PR adds the `interactive` keyword to `plot_run`. This can be used in tests or scripts to non-interactively return a `matplotlib.Figure` object that can be further manipulated or tested upon. This avoids opening a gazillion plot windows whenever `plot_run` is used in loops.

Also adds the option to test multiple data-files in unit testing. To be discussed if _all_ datafiles need to be tested, or a subset.

One TODO to be solved with maintainer/reviewer input.

Prerequisite to #1376 continuation.

### DIFF
--- a/torax/_src/plotting/plotruns_lib_test.py
+++ b/torax/_src/plotting/plotruns_lib_test.py
@@ -58,44 +58,23 @@ class PlotrunsLibTest(parameterized.TestCase):
     test_data_path = os.path.join(test_data_dir, data_file)
     plotruns_lib.load_data(test_data_path)
 
-  @parameterized.product(
-      config_name=[
-          "default_plot_config",
-          "global_params_plot_config",
-          "simple_plot_config",
-          "sources_plot_config",
-      ],
-      data_file=[
-          "test_iterhybrid_rampup.nc",
-      ],
-  )
-  def test_plot_config_smoke_test(self, config_name: str, data_file: str):
-    test_data_dir = paths.test_data_dir()
-    test_data_path = os.path.join(test_data_dir, data_file)
-    config_path = path_utils.torax_path().joinpath(
-        "plotting", "configs", config_name + ".py"
-    )
-    assert config_path.is_file(), f"Path {config_path} is not a file."
-    plot_config = config_loader.import_module(config_path)["PLOT_CONFIG"]
-    fig = plotruns_lib.plot_run(plot_config, test_data_path, interactive=False)
-    assert isinstance(fig, figure.Figure)
-    plt.close(fig)
-
   @parameterized.named_parameters(_generate_all_test_cases())
-  def test_plot_config_all_test(self, config_name: str, data_file: str):
+  def test_plot_config_all(self, config_name: str, data_file: str):
     test_data_dir = paths.test_data_dir()
     config_path = path_utils.torax_path().joinpath(
         "plotting", "configs", config_name + ".py"
     )
-    assert config_path.is_file(), f"Path {config_path} is not a file."
+    self.assertTrue(
+        config_path.is_file(), msg=f"Path {config_path} is not a file."
+    )
     plot_config = config_loader.import_module(config_path)["PLOT_CONFIG"]
     test_data_path = test_data_dir / data_file
     fig = plotruns_lib.plot_run(
         plot_config, str(test_data_path), interactive=False
     )
-    assert isinstance(
-        fig, figure.Figure
-    ), f"Plotting of {test_data_path.name} failed"
+    self.assertIsInstance(
+        fig, figure.Figure, msg=f"Plotting of {test_data_path.name} failed"
+    )
     plt.close(fig)
 
 


torax/_src/plotting/plotruns_lib_test.py
```diff
@@ -29,27 +31,65 @@ class PlotrunsLibTest(parameterized.TestCase):
...
     config_path = path_utils.torax_path().joinpath(
         "plotting", "configs", config_name + ".py"
     )
     assert config_path.is_file(), f"Path {config_path} is not a file."
     plot_config = config_loader.import_module(config_path)["PLOT_CONFIG"]
-    plotruns_lib.plot_run(plot_config, self.test_data_path)
+    data_files = test_data_dir.glob("*.nc")
+    for test_data_path in data_files:
+        # TODO: Allow plotting of runs without 'profiles' and remove exception
+        # or filter out runs without 'profiles' apriori
```
Karel-van-de-Plassche: As of now, this function only plots profiles. As such, would a datafile without the `profiles` key need to be handled by `plot_run`, or should we assume the given datafile always contains `profiles`? Right now fails expectedly with

``` python
_________________________________ PlotrunsLibTest.test_plot_config_all_test2 __________________________________
[gw2] linux -- Python 3.11.13 /home/karel/venvs/toraxvenv/bin/python3.11

self = <plotruns_lib_test.PlotrunsLibTest testMethod=test_plot_config_all_test2>
config_name = 'simple_plot_config'

    @parameterized.product(
    config_name=[
        "default_plot_config",
        "global_params_plot_config",
        "simple_plot_config",
        "sources_plot_config",
    ])
    def test_plot_config_all_test(self, config_name: str):
      test_data_dir = paths.test_data_dir()
      config_path = path_utils.torax_path().joinpath(
          "plotting", "configs", config_name + ".py"
      )
      assert config_path.is_file(), f"Path {config_path} is not a file."
      plot_config = config_loader.import_module(config_path)["PLOT_CONFIG"]
      data_files = test_data_dir.glob("*.nc")
      for test_data_path in data_files:
          try:
>             fig = plotruns_lib.plot_run(plot_config, test_data_path, interactive=False)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

torax/_src/plotting/plotruns_lib_test.py:84:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
torax/_src/plotting/plotruns_lib.py:465: in plot_run
    plotdata1 = load_data(outfile)
                ^^^^^^^^^^^^^^^^^^
torax/_src/plotting/plotruns_lib.py:368: in load_data
    profiles_dataset = data_tree.children[output.PROFILES].dataset
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Frozen({'core_profiles': <xarray.DataTree 'core_profiles'>
Group: /core_profiles
    Dimensions:                      ...t64 4kB 0.0 1.054e+06 ... 1.247e+09
        E_cumulative_external      (time) float64 4kB 0.0 5.808e+05 ... 2.542e+08})
key = 'profiles'

    def __getitem__(self, key: K) -> V:
>       return self.mapping[key]
               ^^^^^^^^^^^^^^^^^
E       KeyError: 'profiles'

../../venvs/toraxvenv/lib/python3.11/site-packages/xarray/core/utils.py:502: KeyError
```

Karel-van-de-Plassche: @jcitrin any preference?

tamaranorman: So all of the recent nc files should have the profiles attribute since 1.0 and I would say we are happy to only support plotting of these

Do you know which is the file that it isn't liking as it could be that this is old and should be removed

tamaranorman:
I would also say that I don't think we necessarily want to test them all but maybe pick out a good subset. Using parameterized.product could be a good way of achieving this (maybe the ones used for checking recomputable in sim_test) or similar

torax/_src/plotting/plotruns_lib_test.py
jcitrin: Sorry @Karel-van-de-Plassche for not having answered before. I agree with Tamara on not supporting legacy formats.

I just checked and the offending file that is not up-to-date is `tests/test_data/test_iterhybrid_predictor_corrector_zeffprofile_qualikiz.nc` , which indeed is not part of the standard integration tests due to the length of execution time. However we can still test it for a couple of timesteps. I'll make a new PR replacing that nc file and making a minimal test for it.

For now you can simply exclude that file from `data_files`

Karel-van-de-Plassche: Thanks for your support! This suggestion is implemented in https://github.com/google-deepmind/torax/pull/1706/commits/514001518990c9a9a1f01beea496ba03180b5f6f

From my side this is ready to be merged.

Karel-van-de-Plassche:
For now I'll do a simple skip. I'm working on a bigger PR locally that might need for plotting a good subset. I think plotting all is not a problem, as the plots are very quick, and the tests are skipped in GitHub CI.

torax/_src/plotting/plotruns_lib_test.py
jcitrin: The offending file has been removed

https://github.com/google-deepmind/torax/pull/1732

So you can revert the exclusion (or keep the pattern but keep an empty exclusion list for now) and we'll bring this in

Karel-van-de-Plassche: I'll keep the pattern for future work. Dropped file in 5a050d8451a45c71debc916aca228f9d0e2ef2b2

Karel-van-de-Plassche:
~Seems there are more offending files? https://github.com/Karel-van-de-Plassche/torax/actions/runs/19328666786/job/55286039500#step:7:873~

EDIT testing in parallel using:
```
pytest -vv -n 1 ./torax/_src/plotting/plotruns_lib_test.py 
```
I see I misunderstood @jcitrin comment, it will be fine as soon a we merge ;) Merged with `main` in ecd69f9a5a87173025773a484214e870aba75bec

torax/_src/plotting/plotruns_lib_test.py
```diff
@@ -18,38 +18,77 @@
...
+from matplotlib.figure import Figure
 from torax._src import path_utils
 from torax._src.config import config_loader
 from torax._src.plotting import plotruns_lib
 from torax._src.test_utils import paths
 
 
 class PlotrunsLibTest(parameterized.TestCase):
+  skip_data_files_list = [
+  ]
```
tamaranorman, line 41: Move onto one line and also make a tuple such that it cannot be changed

Also add the type

torax/_src/plotting/plotruns_lib_test.py
tamaranorman: Is there particular work you are envisaging here? It feels like removing would be nicer as really it is mainly just a sign that we have files which are out of date and should be removed.

In general it is making me think that having a single place we declare all the test files and reusing would make this easier to debug as then we can make it parameterizzed on that without repeating that list elsewhere

jcitrin: It was my suggestion to keep the pattern. Indeed we could just remove it. On having a single place for declaring all test files that's good but maybe in a separate PR?

torax/_src/plotting/plotruns_lib_test.py
Karel-van-de-Plassche: I was not sure if we'd need list manipulation later; I agree with the sentiment.

torax/_src/plotting/plotruns_lib_test.py
Karel-van-de-Plassche: > It was my suggestion to keep the pattern. Indeed we could just remove it. On having a single place for declaring all test files that's good but maybe in a separate PR?

Indeed, I also preferred it, it's why I kept it.

> single place for declaring all test files

Just to be clear, are we talking about the _netCDF files_ here? The test files and definition are defined by the toplevel `pytest` call in principle. If we're talking about the netCDF files, I agree, a global location would be better. As of now, I kept everything in the `plotting` test file, so that it was self-isolated. A place we could define this is in [conftest.py](https://github.com/google-deepmind/torax/blob/main/torax/conftest.py), which would be the `pytest` way to do it. I'm not sure if Abseil has any preferred way to do it. I have been trying a few different way, but thinking about variable scoping is important here (unfortunately).

There is one choice to be made, do we want to test plotting on _every file that is dumped in a location_. In that case, we can define a session scope list that just looks for `tests/test_data/*.nc`. If we want to be explicit, we can just define a tuple in `conftest.py` as noted above. See https://docs.pytest.org/en/7.1.x/how-to/fixtures.html for scoping options.

> that's good but maybe in a separate PR

I do not mind doing it here because
1. Before this PR, the other test files are not explicitly used by the plot tests, and as such, defining a new variable cross-breaking stuff would be unlikely
2. The discussion is then linked to the code change explictly

Karel-van-de-Plassche: > Is there particular work you are envisaging here?

Yes, I'd like to make the current plot tools a bit more re-usable (think, Bokeh and/or Jupyter notebooks), and integratable (think, Holoviews or similar). I have a few local branches to that effect, but first I wanted to increase test coverage a bit before diving in too deep. Just FYI, we had some old work in the same train of thought for tools like JINTRAC and IMAS specifically, e.g. jpyplot https://gitlab.com/jintrac/jetto-pythontools/-/blob/master/jetto_tools/results_gui.py?ref_type=heads and friends.

torax/_src/plotting/plotruns_lib_test.py
tamaranorman: It feels weird to allow list manipulation as then depending what order the tests are executed in or where they execute could alter what happens here

But as commented I would prefer this gets removed

torax/_src/plotting/plotruns_lib_test.py
tamaranorman: Sounds good on the files, we can look at doing that from our side or if you want to do in a follow-up

And re the work it was more work that needed this specific skip sorry I had been looking a bit at the general code. I would ideally prefer if we remove the skip as we don't currently need and we can avoid having lots of unused code around unless there is a specific need and it doesn't feel like this will be needed now we have removed the bad file

Karel-van-de-Plassche: I agree! I'll pick this up tomorrow myself, thanks for the review!

torax/_src/plotting/plotruns_lib_test.py
Karel-van-de-Plassche: > It feels weird to allow list manipulation as then depending what order the tests are executed in or where they execute could alter what happens here

It should not, as any list manipulation should be done pytest-multicore-aware. It's just that the collection should be done on a master core, or all cores, and the for-loop should be `async`-aware (or not exist at all).

As the other discussion; let's remove it here and move to a global test var instead.

jcitrin:
@Karel-van-de-Plassche when checking this internally I noticed that the new test was a bit slow, so added a commit that allows full parallelization of the test, plus some other superficial cleanups for internal style, and also removed that empty list for exclusions since I don't expect we'll need it going forward. lmk if OK and we can merge.

Karel-van-de-Plassche:
Awesome, merge agreed.

jcitrin:
Sorry there was one last change from internal review: removing the redundant smoke test following the improved new comprehensive test.

jcitrin:
In case you are confused about why this was "merged" twice, it's due to having two repositories for TORAX, internal and external (GitHub). The internal repository is the source of truth and [copybara](https://github.com/google/copybara) handles consistency between the repos.

Therefore the workflow is:
1. External PR written and reviewed
2. We trigger a copybara import to the internal repo
3. Some (hopefully) further minor changes done in internal review, ideally just style stuff to make linters happy, etc
4. Submission to internal repo triggers a commit to GitHub, that copybara relates to the PR and "merges" the PR (even though the actual commit may have slightly different code to the final state of the PR). Author attribution is maintained in this process.

However sometimes changes in (3) are not always trivial, and for transparency we add extra commits to the PR so the original authors can see and provide feedback. In this case there was one extra commit on top of that which I did last minute internally (Remove redundant test. Use TestCase assertions), which then seemed to trigger a merge conflict and copybara couldn't merge the PR (even though the code was already committed to main following internal submission and the copybara workflow). So I made the codebases consistent in your branch and merged manually - even if it wasn't strictly necessary but it's nice to see a PR "merged" and not closed - which is why the final diff of the merge was 0, but your attribution is maintained from your previous commits in the PR.

Karel-van-de-Plassche:
Thanks for the clear explanation! Maybe this can be added to https://torax.readthedocs.io/en/latest/contributing.html, https://torax.readthedocs.io/en/latest/contribution_tips.html, or https://torax.readthedocs.io/en/latest/roadmap.html?

Issue #1406, opened 2025-08-01
Validate neoclassical transport implementation


theo-brown:
@jcitrin starting an issue to track this rather than it being lost in emails.

Performing a one-for-one comparison between JETTO (NCLASS) and TORAX (Angioni-Sauter) in an ST geometry. 

Config:
[from_jetto.py.txt](https://github.com/user-attachments/files/21546987/from_jetto.py.txt)

(Uploaded with a .txt suffix to get round Github's file upload restrictions. If this config is generally useful we could maybe think about adding it in some sort of `third_party` directory).

jcitrin:
Thanks. And I can confirm that the reason that chi_ion goes to zero on the magnetic axis is a natural consequence of the trapped fraction scaling.

NCLASS has a switch that modifies the near-axis ion heat flux to a model (likely by Shaing) related to potato orbits. Since NCLASS is not open source, I can't see exactly what's done.

However in TORAX we can consider implementing an on-axis approximation for the expected ion-heat-transport, e.g. following [Shaing 1997 ](https://pubs.aip.org/aip/pop/article-abstract/4/3/771/263553/Ion-transport-process-around-magnetic-axis-in?redirectedFrom=fulltext), with some smooth transition between this value and Angioni-Sauter up to the radius of validity (can relate this radius to the poloidal larmor radius), likely around rho_norm~0.2 or so. Will look into this.

For now, we hope to see in a benchmark reasonable agreement (~20%?) between NCLASS and Angioni-Sauter for rhonorm>0.2 or so, if truly using the same input profiles and geometry.

theo-brown:
> NCLASS has a switch that modifies the near-axis ion heat flux to a model (likely by Shaing) related to potato orbits.

Oh neat! I'll ask around to see whether I can get confirmation as to what's done.

Edit: I've been pointed in the direction of this paper [Helander 2000](https://pubs.aip.org/aip/pop/article/7/7/2878/450176/On-neoclassical-transport-near-the-magnetic-axis) which I think disagrees that potato orbits are the cause? Francis also mentioned that he hasn't found the potato switch in NCLASS to make much of a difference. Instead, he suggested that it might be a problem to do with the validity of the model in the Pfirsch-Schlueter regime.

fcasson:
Hi, just found this, feel free to @ mention me next time when needed.

jcitrin:
No worries. At the moment it's pretty clear what we should do. Checklist below for clarity:

[...]

PR #1719, opened 2025-11-06
Add Shaing corrections to Angioni-Sauter model

theo-brown:
Part of #1406.

Currently in draft form for development purposes. Before merging, will need to move many of the calculations to `physics.collisions`.

@fcasson @jcitrin for discussions.

### Sources

Electrons: https://pubs.aip.org/aip/pop/article/4/5/1375/980556/Electron-transport-processes-close-to-magnetic
Ions: https://pubs.aip.org/aip/pop/article/4/3/771/263553/Ion-transport-process-around-magnetic-axis-in


theo-brown:
In tandem with this PR I am producing a closer comparison of STEP-like plasmas in JETTO and TORAX; this is taking a while, so I thought I'd get the feature development under way as well.

Below is a plot of the initial state from a modified `examples/step_flattop_bgb.py` simulation comparing Shaing and Angioni-Sauter contributions for the full radius.

- Once I've got a more comparable JETTO run (requires turning off a bunch of models) I'll add JETTO NCLASS profiles as a reference.
- The chi produced by this model still don't seem to be looking like what we expect from NCLASS - namely, significant peaking in ion transport on the axis.
- However, there is an obvious improvement over Angioni-Sauter in that transport doesn't disappear on-axis.

image
image

I've added a smooth exponential transition between the two models, shown here:
image

theo-brown:
Some notes from a discussion with Sarah Newton (neoclassical expert at UKAEA) and @fcasson

- NEO solves the force balance, solving for viscosities and transport simultaneously 
- Anecdotally, NEO behaves similarly to NCLASS on axis in JET cases simulated in JETTO (chi neo i explodes at the axis)
- NCLASS uses fitted values for the viscosities as functions of collisionality (?), and then computes transport using those viscosities. The fit is not good everywhere.
- The Shaing 1997 model implemented here uses an asymptotic  expansion for viscosity- effectively like evaluating NCLASS once and using that value always
- Angioni-Sauter is a fit to NEO data, but the fit is not good everywhere.

theo-brown:
In integrated scenarios, the difference in value on-axis caused by this model actually makes a minimal difference, because the ion transport is still so tiny. Compare with the effect of adding in a patch to mimic NCLASS behaviour on the `examples/step_flattop_bgb.py` simulation:

(Solid=AS+Shaing, dashed=AS only)
image

(Solid=with on-axis patch, dashed=AS only)
image

(Solid=with on-axis patch, dashed=AS+Shaing)
image

As a recap of why we're interested in this, current holes at the core could be a huge problem for STEP. As you can see by comparing the q profiles, the ion transport in the core has a noticeable effect on the current hole.

torax/_src/neoclassical/transport/angioni_sauter.py
```diff
@@ -595,3 +662,127 @@ def _calculate_Lmn(
...
+    core_profiles: Core profiles object.
+
+  Returns:
+    Neoclassical transport coefficients.
+  """
+  # Aliases for readability
+  m_ion = core_profiles.A_i * constants.CONSTANTS.m_amu
+  Z_ion = core_profiles.Z_i_face
+  q = core_profiles.q_face
+  delta = geometry.elongation_face
```
jcitrin: Nit: I realize that delta is used in the paper, but it's confusing because normally kappa is elongation and delta is triangularity. I recommending using kappa here and making a comment on the deviation from paper notation

torax/_src/neoclassical/transport/angioni_sauter.py
```diff
@@ -595,3 +662,127 @@ def _calculate_Lmn(
...
+      * constants.CONSTANTS.epsilon_0**2
+      * constants.CONSTANTS.m_e ** (1 / 2)
+      * T_e_J ** (3 / 2)
+      / (
+          core_profiles.n_i.face_value()
+          * Z_ion**4
+          * constants.CONSTANTS.q_e**4
+          * ln_Lambda_ei
+      )
+  )
```
jcitrin: I guess these could go into collisions.py

torax/_src/neoclassical/transport/angioni_sauter.py
```diff
@@ -595,3 +662,127 @@ def _calculate_Lmn(
...
+  nu_ei = 1 / tau_ei
+  v_t_ion = jnp.sqrt(2 * T_i_J / m_ion)
+  v_t_electron = jnp.sqrt(2 * T_e_J / constants.CONSTANTS.m_e)
+  Omega_0_ion = (
+      constants.CONSTANTS.q_e * core_profiles.Z_i_face * geometry.B_0 / m_ion
+  )
+  Omega_0_electron = (
+      constants.CONSTANTS.q_e * geometry.B_0 / constants.CONSTANTS.m_e
+  )
+
```
jcitrin, line 700: nit: can make a comment for readability that these are Larmor frequencies

torax/_src/neoclassical/transport/angioni_sauter.py
```diff
@@ -595,3 +662,127 @@ def _calculate_Lmn(
...
+  Returns:
+    Neoclassical transport coefficients.
+  """
+  # Aliases for readability
+  m_ion = core_profiles.A_i * constants.CONSTANTS.m_amu
+  Z_ion = core_profiles.Z_i_face
+  q = core_profiles.q_face
+  delta = geometry.elongation_face
+  F = geometry.F_face  # Note: denoted I in Shaing
+  R = geometry.R_major
```
jcitrin: can use local geometry.R_major_profile_face

torax/_src/neoclassical/transport/angioni_sauter.py
```diff
@@ -595,3 +662,127 @@ def _calculate_Lmn(
...
+
+  # Chi i term
+  # Equation 74, Shaing March 1997
+  chi_i = nu_ii * (F * v_t_ion / Omega_0_ion) ** (7 / 3) * C_1 ** (2 / 3)
+
+  # Chi e term
+  # Equation 31, Shaing May 1997
+  chi_e = (
+      nu_ei * (F * v_t_electron / Omega_0_electron) ** (7 / 3) * C_1 ** (2 / 3)
+  )
```
jcitrin: Note that the Shaing papers has these has chi_i_psi and chi_e_psi, i.e. heat flux in terms of flux^2/s as opposed to m^2/s (the dimensionality checks out: it's T^4 m^2/s). Therefore we need to divide these by (grad(psi))^2, i.e. `core_profiles.psi.face_grad() / geo.rho_b` , as is also done in Angioni Sauter

This will likely significantly help with the "small transport" you're seeing currently, since this term blows up at rho=0, as it should. After all, with the random walk analogy, at rho=0, Bp=0, and the step size is infinity.

Therefore on-axis there should be some kind of assumption made, like a good default for chi_max, or taking dpsi/dr from the first cell point.

The comparison to NCLASS and the default Shaing multipliers will be interesting, post-fix.

torax/_src/neoclassical/transport/angioni_sauter.py
```diff
@@ -39,20 +41,50 @@
...
   """Pydantic model for the Angioni-Sauter neoclassical transport model."""
 
   model_name: Annotated[
       Literal['angioni_sauter'], torax_pydantic.JAX_STATIC
   ] = 'angioni_sauter'
+  use_shaing_correction: bool = False
+  shaing_chi_i_multiplier: float = 0.8
+  shaing_chi_e_multiplier: float = 1.0
+  shaing_D_e_multiplier: float = 1.0
+  shaing_blend_rho_norm_start: float = 0.2
```
jcitrin: This can be calculated based on some typical potato orbit width. Can see if equation 73 in Shaing March 1997 evaluated on-axis gives reasonable numbers. For more control, using a self-consistent evaluated blend radius can be another Boolean option (default True), and when false reverting to what you have now

theo-brown: I'm not entirely confident about the conversion from psi coordinates to r coordinates in Eq 73.

Δψ is the orbit width in psi - is the conversion still going to be Δrho_norm = Δψ/(dψ/drho_norm)? This is also going to be asymptoting to infinity at the axis. In the introduction they state that psi is no longer proportional to r on axis.

jcitrin: I'd evaluate Eq 73 at rho=0 to get a single value. Then, can just interpolate on (core_profiles.psi - core_profiles.psi[0])  and find which rho_norm it corresponds to. Watch out for 2*pi.

theo-brown:
I've given it a crack. Here's the step_flattop_bgb example with and without Shaing corrections, using the self-consistent potato width. In this case, the potato orbit width for electrons is 6e-4 and for ions is ~0.15 (both in rho norm units).

On-axis dpsi/drhon has machine epsilon added to avoid division by zero, and then is clipped at chi max.

Figure_1

torax/_src/neoclassical/transport/angioni_sauter.py
theo-brown: Yes, will move once we're settled that everything is correct!

jcitrin:
The width for ions looks less than 0.15 in the plot?

Also, since the on-axis values of chi is never actually used, maybe for the sake of plot clarity, we can just set chi[0]=chi[1] ?

theo-brown:
The width looks less because even in the Shaing transport the ion neoclassical contribution is very small compared to the turbulent one, so you only really see the effect when it blows up to infinity at the axis.

With Shaing (L), without Shaing (R)
image



Comparison of the TORAX models with JETTO-NCLASS:
image

jcitrin:
Looked into this a bit closer.

I think there's a missing 2*pi term in the [Wb] to [m] conversion of chi.

This is because the canonical momentum used in the Shaing paper is e\psi, which implies that it's actually divided by from [Wb] psi. The canonical momentum term is actually e*R*A , and the relation between A and psi comes from

psi = int (B * dS) = int ( curlA * dS) = int_contour(A * dl) = 2*pi*A*R . So we see that we must divide by 2*pi to get the canonical momentum term as just psi.

So that means that there would be a (2*pi)^2 multiplication factor still needed in the formula, which I think would get us to O(1) of NCLASS, and the extra tuning coefficient default can be set according to that

jcitrin:
A couple of other points for more simplification:

1. Only chi_i really matters here. I think for simplicity we can only modify chi_i with the Shaing term

2. We can maybe forget about the DeltaPsi? The Shaing terms drop off quickly anyway, likely on the same order at DeltaPsi. So maybe we can just do a blend with a given window width and that's it?

3. Regarding the on-axis point, we can make some decisions like spline extrapolation of Shaing for rho_norm < 0.05 or somesuch (should play with it a bit and see), using similar or the same functions used in the StandardGeometryIntermediates post_init

theo-brown:
Great catch on the psi conversion, sorry for not being on top of that! STEP's chi profiles are now looking much more like what we expect from NCLASS. I haven't adjusted the scaling factors at all, need to see how this performs in ITER cases as well.

image

Regarding simplifications:
- Is electron neoclassical transport much smaller in all cases? If so, then yes we can switch to just doing Shaing correction for the ions
- I've dropped the DeltaPsi bit, reverting back to default window width of rho_norm = 0.2
- Currently I've just set chi[0] = chi[1] for simplicity as per your original suggestion. Since it's never used I reckon this or 0 makes most sense, rather than doing splines?

theo-brown:
If this is good enough, I'll add some tests and move a bunch of the calculations to collisions.py. Lmk if you spot any other TODOs in the meantime.

theo-brown:
Any particular requirements on tests for this? The Angioni-Sauter test is currently only a regression test against NEOS values. I can test against reference values from running standalone NCLASS if required, but would rather not if there's a simpler option as I would have to first learn how to run NCLASS.

fcasson:
> Any particular requirements on tests for this? The Angioni-Sauter test is currently only a regression test against NEOS values. I can test against reference values from running standalone NCLASS if required, but would rather not if there's a simpler option as I would have to first learn how to run NCLASS.

I don't see any reason the test should be different to the Angioni-Sauter case.  NCLASS is probably not better than NEOS?

jcitrin:
> Is electron neoclassical transport much smaller in all cases? If so, then yes we can switch to just doing Shaing correction for the ions

Yes, electron neoclassical transport is always small since it is related to the Larmor radius 

> I've dropped the DeltaPsi bit, reverting back to default window width of rho_norm = 0.2

This is fine.

> Currently I've just set chi[0] = chi[1] for simplicity as per your original suggestion. Since it's never used I reckon this or 0 makes most sense, rather than doing splines?

This is fine.

jcitrin:
> Any particular requirements on tests for this? The Angioni-Sauter test is currently only a regression test against NEOS values. I can test against reference values from running standalone NCLASS if required, but would rather not if there's a simpler option as I would have to first learn how to run NCLASS.

The Angioni-Sauter test is just run against Angioni-Sauter values, to check for regressions. It's not against NEOS (which was used to calibrate Angioni-Sauter).

Can do something similar here: can just update the test to run against the combined Angioni-Sauter+Shaing model.

jcitrin:
Profiles look very good for STEP. How does it compare for ITER? Can set the prefactor default to be something that minimizes the difference to NCLASS for the 2 cases.

jcitrin:
#1804 has just been merged. What remains here before we bring it in? Is it just redoing test_step_flattop_bgb with neoclassical transport with the Shaing coefficient calibrated to NCLASS?

theo-brown:
I have found that the changes to j.B have massively reduced the impact of this PR, to the extent that this PR is potentially no longer even worthwhile. Issues with q profile in STEP core were mainly due to j.B, rather than neoclassical. Only real impact of this PR now is to provide slight flattening of the Ti profile on axis.

Regardless, to wrap things up I've done a comparison with NCLASS in ITER and STEP. Default parameters are tuned for ITER with n_rho=100.

Bad news: it doesn't like changing the grid size :(

ITER, n_rho=100
image

STEP, n_rho=25
image

STEP, n_rho=100
image

theo-brown:
One other potential problem to raise is that the convectivity V_neo_e+V_neo_ware_e in the neoclassical transport model is having problems with loading geometry from IMAS, causing V to explode on axis. I found this in both ITER and STEP cases. I've basically disabled neoclassical V in the STEP case as a result.

jcitrin:
> One other potential problem to raise is that the convectivity V_neo_e+V_neo_ware_e in the neoclassical transport model is having problems with loading geometry from IMAS, causing V to explode on axis. I found this in both ITER and STEP cases. I've basically disabled neoclassical V in the STEP case as a result.

The Ware pinch isn't always negligible. Does just setting zero on-axis help for now? That's not a terrible assumption. If so, I recommend doing so

jcitrin:
On the chi comparison with NCLASS, I'm really happy with the chi_i comparison to NCLASS, that looks good. 

The behaviour with increasing n_rho looks fine. The Shaing term diverges to infinity on axis, so the more refined the grid, the more resolution we have on this divergence.

jcitrin:
> Does just setting zero on-axis help for now?

Or copying from index 1

theo-brown:
Note this isn't a problem with the Shaing correction term, but with the original Angioni-Sauter. I'm not quite sure why it didn't appear before, I think there's a gap in the test suite (no other tests hit neoclassical + imas, as far as I can tell). I can make your suggested change to the Ware term here, doing a bulk+axis type correction?

https://github.com/google-deepmind/torax/blob/fbde8768ddb528fb4d32c94597082b7a29409c19/torax/_src/neoclassical/transport/angioni_sauter.py#L265-L275

theo-brown:
Hmm, even with the fix to V_neo_ware_e now chi_neo_e is hitting the limits on axis.

jcitrin:
> Note this isn't a problem with the Shaing correction term, but with the original Angioni-Sauter. I'm not quite sure why it didn't appear before, I think there's a gap in the test suite (no other tests hit neoclassical + imas, as far as I can tell). I can make your suggested change to the Ware term here, doing a bulk+axis type correction?

Yes that sounds good. Can be in a separate small PR. I'll approve this one now

jcitrin:
> Hmm, even with the fix to V_neo_ware_e now chi_neo_e is hitting the limits on axis.

They are fully independent. So I guess chi_neo_e also needs a bulk+axis treatment. If the only issue is on-axis then it frankly doesn't matter much what is done there, it's not used

jcitrin (approved):

jcitrin:
We are going to release a v1.2.0 imminently, if the neoclassical stuff doesn't make it we can do it in a small v1.2.1

theo-brown:
See #1830 and #1831

PR #1743, opened 2025-11-15
Update STEP example to load plasma composition from IDS

theo-brown:
Uses loader from #1619 to set plasma composition from IDS


jcitrin (approved):

jcitrin:
Could this one be synced? I'll then bring it in

PR #1755, opened 2025-11-19
IMAS output core profiles coupling

MateoBell:
Add transfer function and tests to save TORAX profiles to an IMAS core_profiles or plasma_profiles IDS. Can receive inputs from a single or several time slices.
Inputs are chosen such that it should be straightforward to include a method into StateHistory using this function to save the IDS for an entire simulation in a similar way as it is done with `simulation_output_to_xr`.

@Nush395 : I kept the help function to compute the impurity density scaling and charge states of the individual densities on the cell_plus_boundaries grid. We can continue the discussion we had on whereas it is better to keep it like this or to extend the impurity quantities in `post_processed_outputs` to cell_plus_boundaries grid.

### DIFF
--- a/torax/_src/imas_tools/output/core_profiles.py
+++ b/torax/_src/imas_tools/output/core_profiles.py
@@ -1,0 +1,395 @@
+# Copyright 2025 DeepMind Technologies Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Useful functions to save IMAS core_profiles or plasma_profiles IDSs from
+TORAX objects."""
+from collections.abc import Sequence
+import datetime
+
+import imas
+from imas import ids_toplevel
+import jax
+import jaxtyping as jt
+import numpy as np
+from torax._src import array_typing
+from torax._src import constants
+from torax._src import state
+from torax._src.config import build_runtime_params
+from torax._src.config import runtime_params
+from torax._src.geometry import geometry
+from torax._src.geometry import geometry as geometry_module
+from torax._src.output_tools import output
+from torax._src.output_tools import post_processing
+from torax._src.physics import charge_states
+from torax._src.sources import source_profiles
+from torax._src.torax_pydantic import model_config
+
+
+def core_profiles_to_IMAS(
+    runtime_params_provider: build_runtime_params.RuntimeParamsProvider,
+    torax_config: model_config.ToraxConfig,
+    post_processed_outputs: Sequence[post_processing.PostProcessedOutputs],
+    core_profiles: Sequence[state.CoreProfiles],
+    core_sources: Sequence[source_profiles.SourceProfiles],
+    geometry: Sequence[geometry.Geometry],
+    times: array_typing.FloatVector,
+    ids: ids_toplevel.IDSToplevel = None,
+) -> ids_toplevel.IDSToplevel:
+  """Save TORAX profiles into an IMAS core_profiles IDS.
+
+  The output grid for all 1D quantities is the cell_plus_boundaries one. The
+  function can be used to save an entire StateHistory or a single time slice.
+  If you want to use this function programatically and save a single time
+  slice, please make sure the inputs are Sequences.
+
+  Args:
+    runtime_params_provider: TORAX RuntimeParamsProvider to get the names of
+      the ions and main_ion fractions.
+    torax_config: ToraxConfig object to get number of main ions.
+    post_processed_outputs: Sequence of TORAX PostProcessedOutputs objects.
+    core_profiles: Sequence of TORAX CoreProfiles objects.
+    core_sources: Sequence of TORAX SourceProfiles objects.
+    geometry: Sequence of TORAX Geometry objects.
+    times: Time array of the slices to save.
+    ids: Optional IDS object to be filled. Can be either core_profiles or
+      plasma_profiles. Default is an empty core_profiles IDS. Note that both
+      exists currently from Data Dictionary version 4, with plasma_profiles
+      being the merge of core_profiles and edge_profiles.
+
+  Returns:
+    Filled core_profiles or plasma_profiles IDS object.
+    Signs used for currents and magnetic quantities might not be consistent
+    with IMAS COCOS convention. This should be checked once TORAX COCOS will be
+    defined.
+  """
+  if ids is None:
+    ids = imas.IDSFactory().core_profiles()
+
+  _fill_metadata(ids)
+  _fill_global_quantities(
+      ids,
+      torax_config,
+      post_processed_outputs,
+      core_profiles,
+      geometry,
+      times,
+  )
+  _fill_profiles_1d(
+      ids,
+      runtime_params_provider,
+      post_processed_outputs,
+      core_profiles,
+      core_sources,
+      geometry,
+      times,
+  )
+  return ids
+
+
+def _fill_metadata(ids: ids_toplevel.IDSToplevel) -> None:
+  ids.ids_properties.comment = (
+      'IDS built from TORAX simulation output. Grid based on torax cell grid +'
+      ' boundaries.'
+  )
+  ids.ids_properties.homogeneous_time = 1
+  ids.ids_properties.creation_date = datetime.date.today().isoformat()
+  ids.code.name = 'TORAX'
+  ids.code.description = (
+      'TORAX is a differentiable tokamak core transport simulator aimed for'
+      ' fast and accurate forward modelling, pulse-design, trajectory'
+      ' optimization, and controller design workflows.'
+  )
+  ids.code.repository = 'https://github.com/google-deepmind/torax'
+
+
+def _fill_global_quantities(
+    ids: ids_toplevel.IDSToplevel,
+    torax_config: model_config.ToraxConfig,
+    post_processed_outputs: list[post_processing.PostProcessedOutputs],
+    core_profiles: list[state.CoreProfiles],
+    geometry: list[geometry.Geometry],
+    times: array_typing.FloatVector,
+) -> None:
+  ids.time = times
+  ids.vacuum_toroidal_field.r0 = geometry[0].R_major
+  # TODO: Check signs for B_0 and currents once TORAX COCOS will be defined.
+  # Currents sign are flipped due to the difference between TORAX COCOS
+  # convention and IMAS one.
+  ids.vacuum_toroidal_field.b0 = [
+      geometry_slice.B_0 for geometry_slice in geometry
+  ]
+  Ip = []
+  v_loop = []
+  I_bootstrap = []
+  li_3 = []
+  beta_pol = []
+  beta_tor = []
+  beta_tor_norm = []
+  t_e_volume_average = []
+  n_e_volume_average = []
+  for cp_slice in core_profiles:
+    Ip.append(-1 * cp_slice.Ip_profile_face[-1])
+    v_loop.append(cp_slice.v_loop_lcfs)
+  for ppo_slice in post_processed_outputs:
+    I_bootstrap.append(-1 * ppo_slice.I_bootstrap)
+    li_3.append(ppo_slice.li3)
+    beta_pol.append(ppo_slice.beta_pol)
+    beta_tor.append(ppo_slice.beta_tor)
+    beta_tor_norm.append(ppo_slice.beta_N)
+    # Temperatures are converted from keV to eV(IMAS standard unit).
+    t_e_volume_average.append(ppo_slice.T_e_volume_avg * 1e3)
+    n_e_volume_average.append(ppo_slice.n_e_volume_avg)
+  ids.global_quantities.ip = Ip
+  ids.global_quantities.current_bootstrap = I_bootstrap
+  ids.global_quantities.v_loop = v_loop
+  ids.global_quantities.li_3 = li_3
+  ids.global_quantities.beta_pol = beta_pol
+  ids.global_quantities.beta_tor = beta_tor
+  ids.global_quantities.beta_tor_norm = beta_tor_norm
+  ids.global_quantities.t_e_volume_average = t_e_volume_average
+  ids.global_quantities.n_e_volume_average = n_e_volume_average
+  ids.global_quantities.ion_time_slice = times[0]
+  # TODO: Add main_ion information to outputs
+  main_ion = torax_config.plasma_composition.main_ion.keys()
+  impurities = post_processed_outputs[0].impurity_species.keys()
+  num_ions = len(main_ion) + len(impurities)
+  # Resize global_quantities.ion to create its substructure for every ion.
+  ids.global_quantities.ion.resize(num_ions)
+  for ion in range(num_ions):
+    ids.global_quantities.ion[ion].t_i_volume_average.resize(len(times))
+    ids.global_quantities.ion[ion].n_i_volume_average.resize(len(times))
+
+
+def _fill_profiles_1d(
+    ids: ids_toplevel.IDSToplevel,
+    runtime_params_provider: build_runtime_params.RuntimeParamsProvider,
+    post_processed_outputs: list[post_processing.PostProcessedOutputs],
+    core_profiles: list[state.CoreProfiles],
+    core_sources: list[source_profiles.SourceProfiles],
+    geometry: list[geometry.Geometry],
+    times: array_typing.FloatVector,
+) -> None:
+  # Resize profiles_1d to create its substructure for the different slices.
+  ids.profiles_1d.resize(len(times))
+  for i in range(len(times)):
+    t = times[i]
+    geometry_slice = geometry[i]
+    cp_state = core_profiles[i]
+    cs_state = core_sources[i]
+    post_processed_outputs_slice = post_processed_outputs[i]
+    # TODO: Add main_ion information to outputs to avoid use of runtime_params
+    runtime_params = runtime_params_provider(t)
+
+    ids.profiles_1d[i].time = t
+    _fill_grid(ids, i, geometry_slice, cp_state)
+    _fill_currents(ids, i, cp_state, cs_state)
+    # Temperatures are converted from keV to eV(IMAS standard unit).
+    T_i = cp_state.T_i.cell_plus_boundaries() * 1e3
+    n_e = cp_state.n_e.cell_plus_boundaries()
+    n_i = cp_state.n_i.cell_plus_boundaries()
+    n_impurity = cp_state.n_impurity.cell_plus_boundaries()
+    pressure_thermal_i = cp_state.pressure_thermal_i.cell_plus_boundaries()
+    ids.profiles_1d[i].t_i_average = T_i
+    ids.profiles_1d[i].electrons.temperature = (
+        cp_state.T_e.cell_plus_boundaries() * 1e3
+    )
+    ids.profiles_1d[i].electrons.density = n_e
+    ids.profiles_1d[i].electrons.density_thermal = n_e
+    ids.profiles_1d[i].electrons.density_fast = np.zeros(
+        len(ids.profiles_1d[i].grid.rho_tor_norm)
+    )
+    ids.profiles_1d[i].n_i_total_over_n_e = (n_i + n_impurity) / n_e
+    ids.profiles_1d[i].electrons.pressure_thermal = (
+        cp_state.pressure_thermal_e.cell_plus_boundaries()
+    )
+    ids.profiles_1d[i].pressure_ion_total = pressure_thermal_i
+    ids.profiles_1d[i].pressure_thermal = (
+        cp_state.pressure_thermal_total.cell_plus_boundaries()
+    )
+    _fill_ions(
+        ids,
+        i,
+        runtime_params,
+        post_processed_outputs_slice,
+        cp_state,
+        T_i,
+        n_i,
+        n_impurity,
+        pressure_thermal_i,
+    )
+    Z_eff = output.extend_cell_grid_to_boundaries(
+        [cp_state.Z_eff], np.array([cp_state.Z_eff_face])
+    )[0]
+    ids.profiles_1d[i].zeff = Z_eff
+
+
+def _fill_ions(
+    ids: ids_toplevel.IDSToplevel,
+    i: int,
+    runtime_params: runtime_params.RuntimeParams,
+    post_processed_outputs_slice: post_processing.PostProcessedOutputs,
+    cp_state: state.CoreProfiles,
+    T_i: jt.Float[jax.Array, 't* cell+2'],
+    n_i: jt.Float[jax.Array, 't* cell+2'],
+    n_impurity: jt.Float[jax.Array, 't* cell+2'],
+    pressure_thermal_i: jt.Float[jax.Array, 't* cell+2'],
+) -> None:
+  def _fill_main_ion_quantities(
+      i,
+      ion,
+      symbol,
+      frac,
+      T_i,
+      n_i,
+  ) -> None:
+    ion_properties = constants.ION_PROPERTIES_DICT[symbol]
+    # TODO(b/459479939): i/539) - Indicate supported dd_versions and switch on
+    # that instead of using a try-except.
+    try:
+      ids.profiles_1d[i].ion[ion].name = symbol
+    except AttributeError:
+      # Case ids is plasma_profiles in early DDv4 releases.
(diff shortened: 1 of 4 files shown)


torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,437 @@
...
+from torax._src.config import build_runtime_params
+from torax._src.config import runtime_params
+from torax._src.geometry import geometry
+from torax._src.geometry import geometry as geometry_module
+from torax._src.output_tools import post_processing
+from torax._src.physics import charge_states
+from torax._src.sources import source_profiles
+
+
+def core_profiles_to_IMAS(
```
Nush395, line 39: Would it be possible to break this function down into some smaller helper calls as the whole function is quite large. This is mainly to help with readability.

Maybe breaking down into a function each for:
- global quantities
- profiles
- main ions
- etc.

could help to structure this a bit. wdyt?

MateoBell: yes completely ! I restructured it into coherent chunks. There is just some fields inside profiles_1d that I could not group. 
Let me know if you think this looks better like this or if it still needs to be restructured 😄

MateoBell:
Restructured to use helper calls with the following structure: 

- metadata
- global_quantities
- profiles_1d
   - grid 
   - ions quantities
      - man_ions
      - impurities
   - currents
   - other fields of profiles_1d

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+      post_processed_outputs,
+      core_profiles,
+      core_sources,
+      geometry,
+      times,
+  )
+  return ids
+
+
+def _calculate_impurity_density_scaling_and_charge_states(
```
Nush395: a micronit on code structure: Could you put the `_fill_metadata`, `_fill_global_quantities` and `_fill_profiles_1d` functions before this one. (So that a reader of the code sees those higher level functions first).

Also putting the public function `core_profiles_to_IMAS` at the top as you have done is great!

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+from torax._src.sources import source_profiles
+
+
+def core_profiles_to_IMAS(
+    runtime_params_provider: build_runtime_params.RuntimeParamsProvider,
+    post_processed_outputs: (
+        post_processing.PostProcessedOutputs
+        | list[post_processing.PostProcessedOutputs]
+    ),
+    core_profiles: state.CoreProfiles | list[state.CoreProfiles],
```
Nush395: it feels a bit unnecessarily complex to accept both a list and a single item here. Which do you think is the more common use case? I'm wondering if we can just restrict this function only accept the `Sequence` rather than a single item. 

I think that would:
- make the interface less confusing for users
- make the logic of the function simpler too

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+    core_sources: (
+        source_profiles.SourceProfiles | list[source_profiles.SourceProfiles]
+    ),
+    geometry: geometry.Geometry | list[geometry.Geometry],
+    times: float | array_typing.FloatVector,
+    ids: ids_toplevel.IDSToplevel = imas.IDSFactory().core_profiles(),
```
Nush395: mutable objects like this shouldn't be made as default arguments https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments. Probably best to use `None` instead and create an object inside the function if nothing is provided.

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+    core_profiles: list[state.CoreProfiles],
+    geometry: list[geometry.Geometry],
+    times: array_typing.FloatVector,
+) -> None:
+  ids.time = times
+  ids.vacuum_toroidal_field.r0 = geometry[0].R_major
+  ids.vacuum_toroidal_field.b0 = [
+      geometry_slice.B_0 for geometry_slice in geometry
+  ]  # TODO: Check sign(s) once TORAX COCOS will be defined.
+  ids.global_quantities.ip = [
```
Nush395: I would iterate over `core_profiles` and `post_processed_outputs` just once here (rather than using multiple list comprehensions) and fill in the relevant fields from each

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+  ids.global_quantities.beta_pol = [
+      ppo_slice.beta_pol for ppo_slice in post_processed_outputs
+  ]
+  ids.global_quantities.beta_tor = [
+      ppo_slice.beta_tor for ppo_slice in post_processed_outputs
+  ]
+  ids.global_quantities.beta_tor_norm = [
+      ppo_slice.beta_N for ppo_slice in post_processed_outputs
+  ]
+  # Temperatures are converted from eV (IMAS standard unit) to keV.
```
Nush395: "from keV to eV" ?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+  ids.global_quantities.n_e_volume_average = [
+      ppo_slice.n_e_volume_avg for ppo_slice in post_processed_outputs
+  ]
+  ids.global_quantities.ion_time_slice = times[0]
+  main_ion = runtime_params_provider(
+      times[0]
+  ).plasma_composition.main_ion.fractions
+  impurities = runtime_params_provider(
+      times[0]
+  ).plasma_composition.impurity_names
```
Nush395: I think this can be accessed instead from the post_processed_outputs we have `post_processed_outputs[0].impurity_species.keys()` ?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+  ]
+  # Temperatures are converted from eV (IMAS standard unit) to keV.
+  ids.global_quantities.t_e_volume_average = [
+      ppo_slice.T_e_volume_avg * 1e3 for ppo_slice in post_processed_outputs
+  ]
+  ids.global_quantities.n_e_volume_average = [
+      ppo_slice.n_e_volume_avg for ppo_slice in post_processed_outputs
+  ]
+  ids.global_quantities.ion_time_slice = times[0]
+  main_ion = runtime_params_provider(
```
Nush395: I wonder if this should also be accessed from an output, although I'm not sure it exists atm

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+def _fill_profiles_1d(
+    ids: ids_toplevel.IDSToplevel,
+    runtime_params_provider: build_runtime_params.RuntimeParamsProvider,
+    post_processed_outputs: list[post_processing.PostProcessedOutputs],
+    core_profiles: list[state.CoreProfiles],
+    core_sources: list[source_profiles.SourceProfiles],
+    geometry: list[geometry.Geometry],
+    times: array_typing.FloatVector,
+) -> None:
+  ids.profiles_1d.resize(len(times))
```
Nush395, line 184: what does this logic do?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+
+    ids.profiles_1d[i].time = t
+    _fill_grid(ids, i, geometry_slice, cp_state)
+    _fill_ions(ids, i, runtime_params, post_processed_outputs_slice, cp_state)
+    _fill_currents(ids, i, post_processed_outputs_slice, cp_state, cs_state)
+    ids.profiles_1d[i].electrons.temperature = (
+        cp_state.T_e.cell_plus_boundaries() * 1e3
+    )
+    ids.profiles_1d[i].electrons.density = cp_state.n_e.cell_plus_boundaries()
+    ids.profiles_1d[i].electrons.density_thermal = (
```
Nush395: what is the distinction between this and the above `density` field?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+
+def _fill_ions(
+    ids: ids_toplevel.IDSToplevel,
+    i: int,
+    runtime_params: runtime_params.RuntimeParams,
+    post_processed_outputs_slice: post_processing.PostProcessedOutputs,
+    cp_state: state.CoreProfiles,
+) -> None:
+  def _fill_main_ion_quantities() -> None:
+    ion_properties = constants.ION_PROPERTIES_DICT[symbol]
```
Nush395, line 256: where is this `symbol` defined?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+    ids.profiles_1d[i].ion[index].element[0].a = ion_properties.A
+    ids.profiles_1d[i].ion[index].element[0].z_n = ion_properties.Z
+
+    ids.global_quantities.ion[index].t_i_volume_average[i] = (
+        post_processed_outputs_slice.T_i_volume_avg * 1e3
+    )  # Volume average Ti and ni only available for main ion.
+
+  main_ion = runtime_params.plasma_composition.main_ion.fractions
+  impurity_symbols = runtime_params.plasma_composition.impurity_names
+  impurity_fractions_arr = jnp.stack(
```
Nush395: this will produce a `jax.Array` which we've found to be slow when not under `jax.jit`. Can you use numpy in this module for speed and consistency with the rest of the file?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,497 @@
...
+    try:
+      ids.profiles_1d[i].ion[index].name = symbol
+    except AttributeError:
+      # TODO(b/459479939): i/539) - Indicate supported dd_versions and switch on
+      # that instead of using a try-except.
+      # Case ids is plasma_profiles in early DDv4 releases.
+      ids.profiles_1d[i].ion[index].label = symbol
+    ids.profiles_1d[i].ion[
+        index
+    ].temperature = cp_state.T_i.cell_plus_boundaries()
```
Nush395: this quantity was already computed for `_fill_ion`, is it necessary to extract and save  the same info twice? Similar comment elsewhere

MateoBell: I would say it's good to fill the temperature (or other quantity) for each individual ions even if it is the same. I think it can be useful for compatibility with other codes if you need to extract quantities for a specific ion to just go to its associated `profiles_1d.ion` substructure. Wdyt ?

torax/_src/imas_tools/output/core_profiles.py
MateoBell: From DD description (https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/core_profiles.html#ids-core_profiles): 
density = density thermal + non_thermal (fast). 
@jcitrin  told me there is not fast particles yet in TORAX so I assume we can all consider them to be thermal, for now. Is it correct ?

torax/_src/imas_tools/output/core_profiles.py
MateoBell: This creates the substructure of profiles_1d so we can access using profiles_1d[i]. Without doing this it will give an  `Index error: list index out of range`

torax/_src/imas_tools/output/core_profiles.py
MateoBell: I don't see any output for this, but if it exists I can make the change !

torax/_src/imas_tools/output/core_profiles.py
MateoBell: I think the two main cases will be: 
- Fill core_profiles at the end of a simulation with multiple time steps from a StateHistory object: then it will natively receive a list 
- Fill a single time slice at the end of a step (inside the Muscle3 actor for example), so receive a single item

I agree it would be less confusing and logic lighter, but I didn't know what was the better solution to handle both cases, maybe we can just accept a Sequence, and therefore  just assume we will provide directly a sequence containing a single item if we want to output a single timeslice ?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,499 @@
...
+  impurity_density_scaling, Z_avg_per_species = (
+      _calculate_impurity_density_scaling_and_charge_states(
+          cp_state, runtime_params
+      )
+  )
+  num_of_main_ions = len(main_ion)
+  num_ions = num_of_main_ions + len(impurities)
+  ids.profiles_1d[i].ion.resize(num_ions)
+  # Fill main ions quantities
+  for ion, (symbol, frac) in enumerate(main_ion.items()):
```
MateoBell, line 326: > where is this `symbol` defined?

It is defined here in the for loop inside which the function is called. Is it okay to do it like this ?

torax/_src/imas_tools/output/core_profiles.py
jcitrin: Correct, all thermal for now

torax/_src/imas_tools/output/core_profiles.py
Nush395: if you think passing in a sequence of length 1 from the Muscle3 side is acceptable I think that would be a nicer solution.

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+  The output grid for all 1D quantities is the cell_plus_boundaries one. Values
+  not available on boundaries are copied from neghbouring points. The function
+  can be used to save an entire StateHistory or a single time slice. If you
+  want to use this function programatically and save a single time slice,
+  please make sure the inputs are Sequences.
+
+  Args:
+    runtime_params_provider: TORAX RuntimeParamsProvider to get the names of
+      the ions and main_ion fractions.
+    post_processed_outputs: Single occurence or list of TORAX
```
Nush395: these docstrings just need a little refresh now

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    core_profiles: list[state.CoreProfiles],
+    geometry: list[geometry.Geometry],
+    times: array_typing.FloatVector,
+) -> None:
+  ids.time = times
+  ids.vacuum_toroidal_field.r0 = geometry[0].R_major
+  ids.vacuum_toroidal_field.b0 = [
+      geometry_slice.B_0 for geometry_slice in geometry
+  ]  # TODO: Check sign(s) once TORAX COCOS will be defined.
+  ip = []
```
Nush395: nit: can we call this capitalised `Ip` for consistency with the rest of the codebase

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+  v_loop = []
+  I_bs = []
```
Nush395: is this bootstrap current? Could we call it `I_bootstrap` instead?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+  li_3 = []
+  beta_pol = []
+  beta_tor = []
+  beta_tor_norm = []
+  t_e_volume_average = []
+  n_e_volume_average = []
+  for cp_slice in core_profiles:
+    ip.append(-1 * cp_slice.Ip_profile_face[-1])
```
Nush395: is this because of a specific COCOS?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    v_loop.append(cp_slice.v_loop_lcfs)
+  for ppo_slice in post_processed_outputs:
+    I_bs.append(-1 * ppo_slice.I_bootstrap)
```
Nush395: same here as above re. COCOS

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    t_e_volume_average.append(ppo_slice.T_e_volume_avg * 1e3)
+    n_e_volume_average.append(ppo_slice.n_e_volume_avg)
+  ids.global_quantities.ip = ip
+  ids.global_quantities.current_bootstrap = I_bs
+  ids.global_quantities.v_loop = v_loop
+  ids.global_quantities.li_3 = li_3
+  ids.global_quantities.beta_pol = beta_pol
+  ids.global_quantities.beta_tor = beta_tor
+  ids.global_quantities.beta_tor_norm = beta_tor_norm
+  # Temperatures are converted from keV to eV(IMAS standard unit).
```
Nush395: can you move this comment to line 140 now

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+  ids.global_quantities.t_e_volume_average = t_e_volume_average
+  ids.global_quantities.n_e_volume_average = n_e_volume_average
+  ids.global_quantities.ion_time_slice = times[0]
```
Nush395, line 162: why is only the first time extracted here?

torax/_src/imas_tools/output/core_profiles.py
Nush395: indeed it's not output atm sadly..

I think better than needing to build the provider (which is slightly costly as it requires doing interpolations and stuff) can we access this from the `ToraxConfig` instead?

`torax_config.plasma_composition.main_ion.keys()`

this will require piping in the `ToraxConfig` to this function as well but I think that would be preferable to having to build the provider.

And also could you add a TODO to add this to outputs (just so that we are reminded to do so)

torax/_src/imas_tools/output/core_profiles.py
Nush395: ah great, could you add this as a comment?

torax/_src/imas_tools/output/core_profiles.py
Nush395: and maybe a dumb question but why does the "global quantites" section have a time field?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    cs_state = core_sources[i]
+    post_processed_outputs_slice = post_processed_outputs[i]
+    runtime_params = runtime_params_provider(t)
+
+    ids.profiles_1d[i].time = t
+    _fill_grid(ids, i, geometry_slice, cp_state)
+    _fill_ions(ids, i, runtime_params, post_processed_outputs_slice, cp_state)
+    _fill_currents(ids, i, post_processed_outputs_slice, cp_state, cs_state)
+    ids.profiles_1d[i].electrons.temperature = (
+        cp_state.T_e.cell_plus_boundaries() * 1e3
```
Nush395, line 205: can we have the same comment about converting keV->eV here

torax/_src/imas_tools/output/core_profiles.py
Nush395: I see, thanks!

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    )
+    ids.profiles_1d[i].electrons.density = cp_state.n_e.cell_plus_boundaries()
```
Nush395: just a general comment here for all the `cell_plus_boundaries` quantities. Can we extract them just once and then populate them using that rather than recomputing each time.

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+      [core_profiles.Z_impurity_face[-1]],
+  ])
+  impurity_density_scaling = Z_impurity / Z_avg
+  return impurity_density_scaling, charge_state_info.Z_per_species
+
+
+def _fill_ions(
+    ids: ids_toplevel.IDSToplevel,
+    i: int,
+    runtime_params: runtime_params.RuntimeParams,
```
Nush395, line 240: from what I see the `runtime_params` is not used here, can we avoid computing it and remove from the signature?

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    ids.profiles_1d[i].ion[ion].density_thermal = (
+        cp_state.n_i.cell_plus_boundaries() * frac
+    )
+    ids.profiles_1d[i].ion[ion].density_fast = np.zeros(
+        len(ids.profiles_1d[i].grid.rho_tor_norm)
+    )
+    # Proportion of this ion for pressure ratio computation.
+    total_ions_mixture_fraction = (
+        frac
+        * cp_state.n_i.cell_plus_boundaries()
```
Nush395: Can we compute `cp_state.n_i.cell_plus_boundaries()` just once and use the computed value in the various places in this function

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,482 @@
...
+    runtime_params: runtime_params.RuntimeParams,
+    post_processed_outputs_slice: post_processing.PostProcessedOutputs,
+    cp_state: state.CoreProfiles,
+) -> None:
+  def _fill_main_ion_quantities() -> None:
+    ion_properties = constants.ION_PROPERTIES_DICT[symbol]
+    # Temporary if, will be removed in future versions where the path
+    # should be the same for both core and plasma_profiles.
+    try:
+      ids.profiles_1d[i].ion[ion].name = symbol
```
Nush395, line 260: I also don't see where the `ion` variable is defined for this function (but might be missing something here!)

torax/_src/imas_tools/output/core_profiles.py
Nush395: ahh ok, I think it would be better to pass it directly into the various functions `_fill_main_ion_quantities` and `_fill_impurity_quantities` and add it to those functions arguments

torax/_src/imas_tools/output/core_profiles.py
MateoBell: It is used line 369-70 to extract the ion fractions if Im correct.

torax/_src/imas_tools/output/core_profiles.py
MateoBell: from DD doc: 
 Time slice of the profiles_1d array used to define the ion composition of the` global_quantities/ion` array.
-> It's a reference for the ion composition used in `global_quantities`.

For the time field do you mean this specific `ion_time_slice` or the fact that quantities are 1D array which depends on time ? If the latter, its meant as global quantity derived from the profiles at the different time slices.

torax/_src/imas_tools/output/core_profiles.py
MateoBell: yes ! Added a TODO to check once TORAX convention will be defined as a comment (grouped for B_0, Ip and bootstrap)

torax/_src/imas_tools/output/core_profiles.py
MateoBell: for `_fill_main_ion_quantities` I use the `runtime_params.plasma_composition.main_ion.fractions`. Do you think we can get these fractions elsewhere for the time slice that we are filling ? Otherwise I think I need the provider anyways, but I can replace it inside `_fill_global_quantities` by using `ToraxConfig`, it will at least avoid one build of the provider

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,400 @@
...
+    cp_state: state.CoreProfiles,
+) -> None:
+  ids.profiles_1d[i].grid.rho_tor_norm = np.concatenate(
+      [[0.0], geometry_slice.rho_norm, [1.0]]
+  )
+  ids.profiles_1d[i].grid.rho_tor = np.concatenate(
+      [[0.0], geometry_slice.rho, [geometry_slice.rho_b]]
+  )
+  ids.profiles_1d[i].grid.psi = cp_state.psi.cell_plus_boundaries()
+  ids.profiles_1d[i].grid.psi_magnetic_axis = cp_state.psi._left_face_value()[0]
```
Nush395: we should make these non private methods (remove the underscore) before accessing them outside of that object

torax/_src/imas_tools/output/core_profiles.py
```diff
@@ -0,0 +1,400 @@
...
+
+def _fill_currents(
+    ids: ids_toplevel.IDSToplevel,
+    i: int,
+    cp_state: state.CoreProfiles,
+    cs_state: source_profiles.SourceProfiles,
+) -> None:
+  q_cell = geometry_module.face_to_cell(cp_state.q_face)
+  s_cell = geometry_module.face_to_cell(cp_state.s_face)
+  ids.profiles_1d[i].q = np.concatenate(
```
Nush395: there is a function in torax._src.output_tools.output.py `_extend_cell_grid_to_boundaries` that we can use for all of these extensions (it's cleaner given this functionality already exists).

You should also rename that function to be non private.

Nush395 (commented):
thanks for all the work Mateo, almost there now, have left some final comments.

PR #1791, opened 2025-11-29
fix(geometry): add validation for negative radii in build_circular_geometry

### LINKED ISSUE #1790: [Enhancement] Missing input validation for negative radii in geometry builder
### Title: `[Robustness] Add input validation for non-positive radii in build_circular_geometry`

Currently, the `build_circular_geometry` function allows the initialization of `Geometry` objects with physically impossible parameters, specifically negative or zero values for `R_major` (major radius) and `a_minor` (minor radius).

Because Torax relies on JAX for high-performance simulation, these invalid values are not caught immediately. Instead, they propagate into the physics kernel. This can lead to:

1.  **Silent Failures:** The simulation runs but produces garbage data (e.g., negative volumes).
2.  **Cryptic JAX Errors:** Downstream calculations (like `sqrt` or logarithms) fail during JIT compilation or execution, producing `NaN`s or complex number warnings that are difficult to trace back to the source geometry.

**Solution**
I propose adding an explicit guard clause at the entry point of `torax._src.geometry.circular_geometry.build_circular_geometry`.

The function should validate that:

  * `R_major > 0`
  * `a_minor > 0`

If these conditions are not met, it should raise a `ValueError` immediately (before JAX compilation begins), providing the user with a clear indication of the configuration error.

**Alternative: **
One could rely on assertions within the JAX kernel (`jax.check_tracer_error`), but this adds runtime overhead and complexity. A Python-level check at the builder stage is cleaner and has zero performance cost during the actual simulation.

**Additional context**
I have already implemented this fix locally and verified it with unit tests. I would be happy to open a Pull Request to resolve this.

**Reproduction Snippet:**

```python
from torax.sources import circular_geometry

# This currently returns a valid object instead of raising an error
geo = circular_geometry.build_circular_geometry(
    n_rho=10,
    elongation_LCFS=1.0,
    R_major=-5.0,  # Physically impossible
    a_minor=1.0,
    B_0=1.0,
    hires_factor=1
)
```
**Type: Enhancement / Bug**

geeky33:
## Summary
This PR adds input validation to the `build_circular_geometry` function to strictly enforce positive values for `R_major` and `a_minor`.

## Context / Problem
Previously, the builder accepted negative or zero radii. These invalid physical parameters would propagate into the physics kernel, causing silent failures or cryptic JAX compilation errors (e.g., `NaN`s in downstream calculations) that were difficult to debug.

## Changes
- **Source:** Added a guard clause in `torax/_src/geometry/circular_geometry.py` to raise a `ValueError` immediately if `R_major <= 0` or `a_minor <= 0`.
- **Tests:** Added a new test case `test_build_circular_geometry_negative_radius_raises` in `circular_geometry_test.py` to verify that both invalid major and minor radii are caught correctly.

## Fixes
Fixes #1790 

## Testing Plan
- [x] Verified unit tests locally: `pytest torax/_src/geometry/tests/circular_geometry_test.py` (**Passed**)
- [x] Verified full simulation suite locally (subset): `pytest torax/tests/sim_test.py` (**Passed**)

### DIFF
--- a/torax/_src/geometry/circular_geometry.py
+++ b/torax/_src/geometry/circular_geometry.py
@@ -87,6 +87,10 @@ def build_circular_geometry(
   Returns:
     A Geometry instance.
   """
+  if R_major <= 0:
+    raise ValueError(f"R_major must be positive, got {R_major}")
+  if a_minor <= 0:
+    raise ValueError(f"a_minor must be positive, got {a_minor}")
   # circular geometry assumption of r/a_minor = rho_norm, the normalized
   # toroidal flux coordinate.
   # Define mesh (Slab Uniform 1D with Jacobian = 1)
--- a/torax/_src/geometry/tests/circular_geometry_test.py
+++ b/torax/_src/geometry/tests/circular_geometry_test.py
@@ -63,7 +63,29 @@ class CircularGeometryTest(absltest.TestCase):
     )
     # Make sure you can call the function with geo as an arg.
     foo(geo)
+  def test_build_circular_geometry_negative_radius_raises(self):
+    """Checks that negative radii raise a ValueError."""
+    # Test negative major radius
+    with self.assertRaisesRegex(ValueError, "R_major must be positive"):
+      circular_geometry.build_circular_geometry(
+          n_rho=10,
+          elongation_LCFS=1.0,
+          R_major=-5.0,
+          a_minor=1.0,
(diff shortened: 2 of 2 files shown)


Nush395:
As discussed in the issue I think this change is unfortunately not needed but thanks for the contribution regardless!

PR #1806, opened 2025-12-03
Change expected_impurities to excluded_impurities in plasma_composition_from_IMAS

### LINKED ISSUE #1789: IMAS plasma composition surprising behaviour
@theo-brown , @MateoBell 

In #1775 I "innocently" added He as a new `constants.IonProperty`, as an alias to He4.

This significantly modified the `step_flattop_bgb` result. Looking deeper, it's because `He` is in the IDS and is now being picked up as an impurity, whereas before it wasn't.

1. @theo-brown , is having all 3 of Xe, Ar, He the desired impurity mix for this STEP case? If so then I can update the nc file in this PR. Note that the ~10% He made a very significant impact to the scenario, around ~50% different temperatures. It's probably good practice to use the optional 'expected_impurities' argument to avoid surprises

2. @MateoBell , thinking again, maybe it's a bit of a footgun to only parse impurities that are in `constants.ION_PROPERTIES_DICT.keys()`, and ignore others . This can lead to surprises like we see here, that impurities that the user thinks should be valid, are ignored. Maybe if an unidentified impurity is found, we throw an error, and then instead of "expected_impurities" we can have an "excluded_impurities" so the user can choose to filter out impurities that maybe don't matter so much but TORAX can't handle (if these exist). WDYT?

MateoBell:
Closes #1789 
Parse all ions by default to avoid ignoring ions without the user knowing.

Changes: 

- Rename `expected_impurities`  to `excluded_impurities` arg. Now parse all ions in the IDS except those in `excluded_impurities`. 
- Add `_validate_ids_ions`  helper function to validate that all parsed ions have recognized TORAX symbols and raise informative error if not 
- Updated tests to match new behavior


jcitrin (approved):

jcitrin:
@Nush395 could you also review this one please?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -221,19 +224,35 @@ def _get_time_and_radial_arrays(
...
 ) -> None:
-  """Checks that the expected_ions symbols are in the ids_ions."""
-  for ion in expected_ions:
-    if ion not in constants.ION_PROPERTIES_DICT.keys():
-      raise (KeyError(f"{ion} is not a valid symbol of a TORAX valid ion."))
-    if ion not in ids_ions:
+  """Check if all ions are recognized and expected_main_ions present in the IDS."""
+  # Check if IDS ion symbol is valid.
+  for ion in parsed_ions:
+    if str(ion) not in constants.ION_PROPERTIES_DICT.keys():
```
Nush395: why is the `str` necessary here? According to the type this should already be a string.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -221,19 +224,35 @@ def _get_time_and_radial_arrays(
...
       raise (
-          ValueError(
-              f"The expected ion {ion} cannot be found in the input"
-              " IDS or has no valid data. \n Please check that the IDS is"
-              " properly filled"
+          KeyError(
+              f"{ion} is present in the IDS but not a valid symbol of a TORAX"
```
Nush395: Should this read "{ion} is present in the IDS but not a valid TORAX ion." ?

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -221,19 +224,35 @@ def _get_time_and_radial_arrays(
...
+              " ion. Check typing or add the ion to the excluded_impurities."
           )
       )
+  # Check presence of main_ion_symbols in the IDS if not default value.
+  if main_ion_symbols is not constants.HYDROGENIC_IONS:
```
Nush395: `is` compared the actual identity of the `main_ion_symbols` against the `constants.HYDROGENIC_IONS` which will only be `True` if the default arg above is used or passed in.

Is that intended? Maybe you want to use `==` here or check for a subset instead?

torax/_src/imas_tools/input/core_profiles.py
MateoBell: If i dont use str I get "TypeError: unhashable type: 'IDSString0D'". Don't know if it' s the best way to solve the issue.

torax/_src/imas_tools/input/core_profiles.py
MateoBell: Yes it was intended. I was thinking that action to provide a non-default value for expected_main_ions triggers validation as it implies that the user specifies which ions should be present in the IDS. So it would make sense to check that user expected ions are indeed present in these case. Wdyt ?

torax/_src/imas_tools/input/core_profiles.py
Nush395: oh interesting, I'd add that as a comment above then explaining why the cast to `str` is done

torax/_src/imas_tools/input/core_profiles.py
Nush395: ah interesting, that's quite clever but a bit non-standard so I'd advise just always doing the validating check (even if it is a bit redundant in the case of using the default arg) as it should be pretty quick.

MateoBell: But then it might fail I think in the `if ion not in parsed_ions:` when value is default ? because not all hydrogenic ions have to be in the ids. For example for DD plasma there will be no T so it would fail the validation if we do not explicitly set the `main_ion_symbols`.

MateoBell: But I think I can replace it with `==`. I don't see a case where user defined expected main ions would be all 3 H, D and T ? So `==` check would be strong enough.

Nush395: I think it's a little confusing to have an exception in the code for the default arg case, we should try have the code not depend on whether the default is used or not I think is my main issue.

Alternatively we could also remove the default so that a user is forced to pick their `main_ion_symbols` and we validate against that.

MateoBell: what if I change default to `None` ? If None we deactivate validation for the main_ions and collect main ion within `constants.HYDROGENIC_IONS`, and otherwise trigger the validation ? 
Still change the behavior when we use default arg or not but maybe it's less confusing this way ? 

I think it's a bit too constraining to force the user to specify `main_ions_symbols` every time, but if you think it's cleaner like this, I'm fine with it.

Nush395: I think that sounds like a good solution, we separate out the configuration from the input validation that way.

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -221,19 +231,37 @@ def _get_time_and_radial_arrays(
...
-              f"The expected ion {ion} cannot be found in the input"
-              " IDS or has no valid data. \n Please check that the IDS is"
-              " properly filled"
+          KeyError(
+              f"{ion} is present in the IDS but not a valid TORAX ion. Check"
+              "typing or add the ion to the excluded_impurities."
           )
       )
+  # Check presence of main_ion_symbols in the IDS if explicitly provided.
+  if validate_main_ions:
```
Nush395: nit: I'd just separate this out into another smaller function that is called in the `else` branch and avoid the need to create a separate `validate_main_ions` flag.

MateoBell:
@Nush395 : Should these new functions go into `validation.py` now ?

Nush395:
> @Nush395 : Should these new functions go into `validation.py` now ?

Sorry for the slow reply. I think moving into `validation.py` would make sense.

MateoBell:
Ok I moved the files and add small unit tests now they are in separate module ! Also rebased onto main and semi-squashed it. Let me know if there is still any change to make

torax/_src/imas_tools/input/core_profiles.py
```diff
@@ -123,8 +123,8 @@ def profile_conditions_from_IMAS(
 def plasma_composition_from_IMAS(
     ids: ids_toplevel.IDSToplevel,
     t_initial: float | None = None,
-    expected_impurities: Collection[str] | None = None,
-    main_ions_symbols: Collection[str] = constants.HYDROGENIC_IONS,
+    excluded_impurities: Collection[str] | None = None,
```
Nush395, line 126: can we make these required keyword arguments (with `*`)?

Reason: I often find for things where there are multiple arguments with the same type that helps make callers more elegant/clear.

Nush395 (approved):

Nush395:
If you squash I can also address the final comments myself in the internal review :) whatever is easiest.

MateoBell:
Just rebased and squashed. If you don't mind to make the changes in the internal review, thanks a lot Anushan ! :)

PR #1807, opened 2025-12-04
Fix issues reproduced on Windows

vanzay:
Fixed small issues reproduced on Windows:
- file system related operations
- changed the command for running python in a subprocess

### DIFF
--- a/torax/tests/persistent_cache_test.py
+++ b/torax/tests/persistent_cache_test.py
@@ -19,6 +19,7 @@ import io
 import os
 import re
 import subprocess
+import sys
 
 from absl.testing import absltest
 from absl.testing import parameterized
@@ -51,7 +52,7 @@ class PersistentCacheTest(parameterized.TestCase):
         torax.__path__[0], 'run_simulation_main.py'
     )
     assert os.path.exists(run_simulation_main_path)
-    command = ['python3', run_simulation_main_path]
+    command = [sys.executable, run_simulation_main_path]
 
     subprocess_args = command + flags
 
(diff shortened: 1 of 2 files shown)


torax/tests/persistent_cache_test.py
```diff
@@ -51,7 +52,7 @@ def test_persistent_cache(self):
         torax.__path__[0], 'run_simulation_main.py'
     )
     assert os.path.exists(run_simulation_main_path)
-    command = ['python3', run_simulation_main_path]
+    command = [sys.executable, run_simulation_main_path]
```
Nush395, line 55: what is the error in Windows that requires this change?

We are unable to use `sys.executable` on our internal systems so would be good to have something that works on all systems here.

vanzay: The Python distribution for Windows doesn't have `python3` command. Instead, it uses just `python`. Therefore, `python3` is an unknown command on Windows.

Also, when I changed the command in the code from `python3` to `python`, the subprocess didn't use the project's virtual environment and threw an exception about missing libraries.

`sys.executable` solved both errors.

I don't know the reason why it can't be used on your internal systems, so I can't offer a workable solution if one exists.
Perhaps, to avoid breaking internal systems, the best solution is to leave a comment on the issue in the code an remove this fix from the PR.

Nush395: I see, no worries, fine to leave as is then.

Nush395 (approved):

Nush395:
For future changes note we recommend adding the "ready-for-review" tag so that the change is flagged to us. https://github.com/google-deepmind/torax/blob/main/CONTRIBUTING.md#code-reviews

PR #1811, opened 2025-12-06
solved IMAS plasma composition surprising behaviour

### LINKED ISSUE #1789: IMAS plasma composition surprising behaviour

paras-chinchalkar:
Fix IMAS plasma composition surprising behaviour (#1789)
Problem
When "He" was added as a new constants.IonProperty (alias to He4), it started being picked up as an impurity from IMAS IDS data, significantly modifying the step_flattop_bgb result (~50% different temperatures). The root cause was that the code silently ignored impurities not in constants.ION_PROPERTIES_DICT.keys(), leading to unexpected behavior when new ions were added to the dictionary.
Solution
This PR addresses the issue by:
Replacing expected_impurities with excluded_impurities: Changed the API to be more intuitive - users can now explicitly exclude impurities they don't want to process, rather than having to specify what they expect.
Adding error handling for unidentified impurities: The function now raises a clear ValueError when an ion is found in the IDS that is not in constants.ION_PROPERTIES_DICT and not in excluded_impurities. This prevents silent failures and makes it explicit when new ions are being processed.
Improved user control: Users can now explicitly control which impurities are processed using the excluded_impurities argument, preventing surprises when new ions are added to ION_PROPERTIES_DICT.
Changes Made
Modified plasma_composition_from_IMAS() in torax/_src/imas_tools/input/core_profiles.py:
Replaced expected_impurities parameter with excluded_impurities
Added validation to raise ValueError for unidentified impurities
Added logic to skip excluded impurities during processing
Updated all tests to use the new excluded_impurities API
Updated documentation strings to reflect the new behavior
Testing
Updated existing tests in torax/_src/imas_tools/input/tests/core_profiles_test.py
Added tests for the new error handling behavior
All linting checks pass
Impact
This change ensures that:
Users are explicitly notified when unidentified impurities are found
Users have control over which impurities are processed
No more silent failures that lead to unexpected simulation results
The API is more intuitive and user-friendly
Fixes #1789

### DIFF
--- a/torax/_src/imas_tools/input/core_profiles.py
+++ b/torax/_src/imas_tools/input/core_profiles.py
@@ -114,7 +114,7 @@ def profile_conditions_from_IMAS(
 def plasma_composition_from_IMAS(
     ids: ids_toplevel.IDSToplevel,
     t_initial: float | None = None,
-    expected_impurities: Collection[str] | None = None,
+    excluded_impurities: Collection[str] | None = None,
     main_ions_symbols: Collection[str] = constants.HYDROGENIC_IONS,
 ) -> Mapping[str, Any]:
   """Returns dict with args for plasma_composition config from a given ids.
@@ -129,7 +129,7 @@ def plasma_composition_from_IMAS(
   Note that if the input ids does not contain info for the different ion
   species, the function will not raise an error but return an empty dict for
   `main_ion` and `species` plasma_composition keys unless the expected species
-  are specified using the expected_impurities and main_ion_symbols args.
+  are specified using the main_ion_symbols arg.
 
   Args:
     ids: A core_profiles IDS object. The IDS can contain multiple time slices.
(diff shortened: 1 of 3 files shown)


jcitrin:
Duplicate. Already tackled in #1806

PR #1831, opened 2025-12-09
Extrapolate chi_neo_e and V_neo_ware_e

### LINKED ISSUE #1830: Extrapolate neoclassical transport values on axis to avoid hitting limits
When loading geo from IMAS (e.g. step_flattop_bgb.py) V_neo_ware_e and chi_neo_e both hit their limits on axis.

The V_neo_ware_e problem can cause simulations to crash  - currently in `step_flattop_bgb.py` this is avoided by effectively disabling neoclassical convection.

theo-brown:
Fixes #1830 

Before, without V_e clamped to 0 in neoclassical - causes sim crash due to unphysical density peaking.
image

Before, with V_e clamped to 0 in neoclassical - note on-axis peak in chi_e (only has a cosmetic effect)
image

After
image


jcitrin (commented):
For consistency in pattern, should do this bulk+axis extrapolation for all the Neoclassical transport outputs.

And then would need to fix the tests (I guess the issue is the unit test + relevant sim_tests that used this)

jcitrin (approved):

jcitrin:
@theo-brown could you please sync this one to main? There will be some merge conflicts due to changes done on top of the Shaing PR following internal review. Thanks! Then we'll bring it in too.

PR #1846, opened 2025-12-12
Fix imas loader test due to new imas-python tag

MateoBell:
New imas-python 2.1.0 release disabled implicit conversion between major DD versions [#82](https://github.com/iterorganization/IMAS-Python/pull/82) which broke imas loader test (change from warning to RuntimeError). 

Updated imas-python requirement and fixed the test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "pydantic>=2.10.5",
     "tqdm>=4.67.0",
     "treelib>=1.3.2",
-    "imas-python>=2.0.1",
+    "imas-python>=2.1.0",
     "typeguard==2.13.3",
 ]
 
--- a/torax/_src/imas_tools/input/tests/loader_test.py
+++ b/torax/_src/imas_tools/input/tests/loader_test.py
@@ -54,7 +54,7 @@ class IMASLoaderTest(parameterized.TestCase):
 
   def test_load_older_dd_version_without_explicit_convert_raises(self):
     directory = pathlib.Path(__file__).parent
-    with self.assertRaises(AttributeError):
+    with self.assertRaises(RuntimeError):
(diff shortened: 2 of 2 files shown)


Nush395 (commented):
2.1.0 isn't released from what I can tell at https://github.com/iterorganization/IMAS-Python. Is this a private version that you are using?

MateoBell:
I think there is just this tag for now https://github.com/iterorganization/IMAS-Python/releases/tag/2.1.0-post1. It triggered a version change for the version use in the pipeline I guess, gave failures when I committed on #1630

Nush395:
Ah interesting, we need to import the library but I think there is a small issue with the release that prevents our tests from running without imas-core that I've raised https://github.com/iterorganization/IMAS-Python/issues/88.

For now I'm proposing restricting to use 2.0.1, is that suitable for your use case?

MateoBell:
Yes it perfectly fine for me I think every feature should work fine with 2.0.1 for now ! 
I was thinking about restricting it to 2.0.1 as well to fix the issue, and can upgrade once this issue is fixed.

Nush395:
> Yes it perfectly fine for me I think every feature should work fine with 2.0.1 for now ! I was thinking about restricting it to 2.0.1 as well to fix the issue, and can upgrade once this issue is fixed.

Awesome, thanks Mateo :)

Nush395:
Hey Mateo, sorry for the wait on this. Just an update, we're looking to add imas_core to our deps so should be able to bring this in soon.

MateoBell:
great happy to hear that ! Don't worry for the wait there is no hurry on this :)

Nush395:
I'm actually going to close this as we'll be forced to make this change internally when we do our upgrade so this PR won't be able to be applied. Thanks for keeping it open for so long.

PR #1852, opened 2025-12-15
Add Redl model of bootstrap current

### LINKED ISSUE #1612: Add Redl model of bootstrap current
Part of ongoing JETTO validation. JETTO/NCLASS shows close agreementh with Redl 2021, but worse agreement with Sauter 1999. Updating the bootstrap current to the Redl model (as noted in the comment) would improve quality of JETTO/TORAX comparisons, particularly in large bootstrap regimes (eg ST).

https://github.com/google-deepmind/torax/blob/0941209539b8a337c8004fc8e35254bf20a459a1/torax/_src/neoclassical/bootstrap_current/sauter.py#L112

theo-brown:
Prompted by JINTRAC NCLASS comparison of Sauter and Redl models, this has been bumped up my priority list!

~~When I have time, I'll try and rearrange some of the definitions and files so there's less duplication.~~

Closes #1612.


theo-brown:
STEP (dashed Redl, solid Sauter)

Figure_1

ITER  (dashed Redl, solid Sauter)

Figure_1

theo-brown:
@jcitrin as this is really an extension of the Sauter model, would you like it as a separate class, subclass, or flag within Sauter?

jcitrin:
Excellent, thanks for this.

It would be good to have redl as a separate class, even if it's quite similar to sauter, and ensure that "sauter" and "redl" can be distinct model names. Possibly a lot from both can be separated out to use the common formulas module and duplication minimized, as you say.

It may be good to do this in more than 1 PR, separating out refactoring and new feature.

lmk when review is necessary and we can get back to this in the new year. Have a great holiday period!

theo-brown:
I suggest the following design:

1. Move `neoclassical.formulas` to a subpackage, so that we can have `neoclassical.formulas.shared`, `neoclassical.formulas.sauter` and `neoclassical.formulas.redl`. 

2. Extract the shared code from https://github.com/google-deepmind/torax/blob/2aeaa4869474d29ba385df21b89cd8cc3c2ea51e/torax/_src/neoclassical/bootstrap_current/sauter.py#L152-L183
into a new function

```python
neoclassical.formulas.shared._calculate_analytic_bootstrap_current(
    *,
    bootstrap_multiplier: float,
    n_e: cell_variable.CellVariable,
    n_i: cell_variable.CellVariable,
    T_e: cell_variable.CellVariable,
    T_i: cell_variable.CellVariable,
    p_e: cell_variable.CellVariable,
    p_i: cell_variable.CellVariable,
    geo: geometry_lib.Geometry,
    L31: array_typing.FloatVectorFace,
    L32: array_typing.FloatVectorFace,
    L34: array_typing.FloatVectorFace,
    alpha: array_typing.FloatVectorFace,
  ) -> base.BootstrapCurrent
```

3. Make a new function for each of Sauter and Redl that computes the terms that are different, e.g.
```python
neoclassical.bootstrap.sauter.calculate_bootstrap_terms(...) -> ...:
  f_trap = neoclassical.formulas.sauter.calculate_ftrap(...)
  log_lambda_ei = collisions.calculate_log_lambda_ei(...)
  log_lambda_ii = collisions.calculate_log_lambda_ii(...)
  nu_e_star = collisions.calculate_nu_e_star(...)
  nu_e_star = collisions.calculate_nu_i_star(...)

  L31 = neoclassical.formulas.sauter.calculate_L31(...)
  L32 = neoclassical.formulas.sauter.calculate_L32(...)
  L34 = neoclassical.formulas.sauter.calculate_L34(...)
  alpha = neoclassical.formulas.sauter.calculate_alpha(...)
 
  return L31, L32, L34, alpha
```

I think this is the most simple option that minimises duplication. Also I think this structure sets up Andreas' intent to do the new conductivity model. @jcitrin thoughts?

jcitrin:
@theo-brown this structure sounds good to me.

@tamaranorman , @Nush395 , any thoughts on the proposed structure? Context: Theo and Andreas Redl are implementing the Redl model for bootstrap current and conductivity, which is an upgrade on the sauter model. Structurally it's very similar to the sauter models, but have slightly different formulas and coefficients. Theo's proposal aims to minimize the duplication and still have both `sauter` and `redl` as model types.

torax/tests/sim_test.py
```diff
@@ -22,18 +22,16 @@
 import os
 from typing import Final, Sequence
 
-from absl.testing import absltest
-from absl.testing import parameterized
 import numpy as np
+import xarray as xr
+from absl.testing import absltest, parameterized
```
tamaranorman: Revert these changes the google style guide recommend everything be imported per module excluding types or similar

torax/_src/neoclassical/bootstrap_current/redl.py
tamaranorman, line 1: Can we add a unit test on the numerics of this file etc instead of just testing through integration tests?

torax/tests/test_data/test_iterhybrid_predictor_corrector_redl.py
tamaranorman, line 1: Do we need this if its also tested as part of the step case? Just that as we add more and more of these the time taken for tests change - it would be cleaner if this could be tested in a unit test

tamaranorman:
Looks good in its current form - one question is when we talk about rearranging what does that look like as this currently is just adding another implementation

theo-brown:
Thanks for your review @tamaranorman! I was mainly @'ing you for your thoughts on the comment, sorry.

See my comment above for my proposed structure ([link](https://github.com/google-deepmind/torax/pull/1852#issuecomment-3699226369)) - I didn't want to waste time implementing before a design had been chosen.

The "changes to structure" are just moving things around, rather than introducing any new classes etc. I think Jonathan was just to check whether there's an alternative (eg create a new superclass that both of these inherit from) or whether having this level of code duplication is ok/acceptable/avoidable.

tamaranorman:
That proposed structure seems reasonable to me - maybe returning a dict that can then be unpacked using ** just as with lots of scalars its worth making sure the return values are clear

torax/_src/neoclassical/bootstrap_current/sauter.py
```diff
@@ -65,10 +65,6 @@ def calculate_bootstrap_current(
         q_face=core_profiles.q_face,
         geo=geometry,
     )
-    return base.BootstrapCurrent(
-        j_parallel_bootstrap=result.j_parallel_bootstrap,
-        j_parallel_bootstrap_face=result.j_parallel_bootstrap_face,
-    )
```
theo-brown: Was there a reason for the original Sauter model reconstructing the BootstrapCurrent object? Does it make type inference easier or something?

torax/_src/neoclassical/bootstrap_current/sauter.py
```diff
@@ -16,15 +16,17 @@
...
 from torax._src import array_typing
 from torax._src import state
 from torax._src.config import runtime_params as runtime_params_lib
 from torax._src.fvm import cell_variable
 from torax._src.geometry import geometry as geometry_lib
-from torax._src.neoclassical import formulas
+from torax._src.neoclassical import formulas as neoclassical_formulas
 from torax._src.neoclassical.bootstrap_current import base
-from torax._src.neoclassical.bootstrap_current import runtime_params as bootstrap_runtime_params
+from torax._src.neoclassical.bootstrap_current import (
```
tamaranorman: Remove this change we only allow one thing to be imported

torax/_src/neoclassical/bootstrap_current/sauter.py
```diff
@@ -16,15 +16,17 @@
 from typing import Annotated, Literal
 
 import jax
-import jax.numpy as jnp
+
```
tamaranorman: remove whitespace

torax/_src/neoclassical/bootstrap_current/sauter.py
tamaranorman: I think this was leftover from when _calculate_bootstrap_current returned a different object

torax/_src/neoclassical/formulas/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 DeepMind Technologies Limited
```
tamaranorman, line 1: Generally we leave the init files empty unless we need to - follow that convention here?

torax/_src/neoclassical/formulas/common.py
```diff
@@ -310,3 +256,55 @@ def calculate_poloidal_velocity(
...
+      + nicoeff * dlnni_drnorm
+      + tecoeff * dlnte_drnorm
+      + ticoeff * dlnti_drnorm
+  )
+  j_parallel_bootstrap = geometry_lib.face_to_cell(j_parallel_bootstrap_face)
+
+  return bootstrap_current_base.BootstrapCurrent(
+      j_parallel_bootstrap=j_parallel_bootstrap,
+      j_parallel_bootstrap_face=j_parallel_bootstrap_face,
+  )
```
tamaranorman: Is this flagging that we need whilespace here?

torax/_src/neoclassical/formulas/tests/sauter_test.py
```diff
@@ -0,0 +1,134 @@
...
+    -0.08055448,
+    -0.06213122,
+    -0.03385067,
+    0.01149523,
+    0.08557197,
+    0.16296924,
+])
+
+if __name__ == '__main__':
+  absltest.main()
```
tamaranorman: New line

torax/_src/neoclassical/formulas/tests/redl_test.py
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+from torax._src.config import build_runtime_params
+from torax._src.core_profiles import initialization
+from torax._src.neoclassical import formulas
+from torax._src.physics import collisions
+from torax._src.torax_pydantic import model_config
+
+# pylint: disable=invalid-name
+
+_N_RHO = 10
```
tamaranorman: Just inline this one?

torax/_src/neoclassical/formulas/tests/sauter_test.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+    np.testing.assert_allclose(L31, _L31_EXPECTED, atol=_A_TOL, rtol=_R_TOL)
+
+  def test_L32_values_are_correct(self):
+    L32 = formulas.sauter.calculate_L32(
+        self.f_trap, self.nu_e_star, self.core_profiles.Z_eff_face
+    )
+    np.testing.assert_allclose(L32, _L32_EXPECTED, atol=_A_TOL, rtol=_R_TOL)
+
+_L31_EXPECTED = np.array([
```
tamaranorman, line 107: If we are using specific values - is there a way we can make the test so that it won't be affected by other tweeks in the code - I am thinking trying to use even less of the code - maybe rngs

theo-brown: Yes, for sure - however, what I have here is currently the old stuff, just moved around not changing. I'd quite like this PR to stay fairly clean if that's ok - maybe add this to my todo list for later?

theo-brown:
Thanks for the review @tamaranorman. Sorry that I still haven't managed to make my linter sync up with the Google one!

theo-brown:
This has been merged, I just need to fix the copybara error.

PR #1926, opened 2026-01-20
Improve radiative model logic (Mavrin) at low temperatures

theo-brown:
Currently WIP. 

Idea is to:
- Use the high-temperature (coronal) Mavrin model where possible
- When the temperature drops to low, switch to using the low-temperature (edge, non-coronal) Mavrin model

Potential later work:
- Do something smart when we drop below the threshold of the non-coronal Mavrin model.
- Move things around so these models are kept in a generic location (e.g. physics?)


theo-brown:
Example cooling rate curves.

image
image
image

PR #1944, opened 2026-01-27
IMAS core_sources: Read ion particle source if electron source is not available

MateoBell:
Closes #1938. 

Waiting for #1918 to be merged to add unit test with example core_sources


tamaranorman:
This seems to now be unblocked, is that correct?

MateoBell:
Yes it should be good now !

Nush395 (approved):
LGTM!

PR #2130, opened 2026-04-22
Update IMAS DD version to 4.1.1

MateoBell:
Upgrade IMAS DD version from 4.0.0 to 4.1.1 and renames changed fields. This remove the need for the` try:.. except` when mapping from/to `plasma_profiles `IDSs. 
Also adds more metadata to the output IDSs

### DIFF
--- a/torax/_src/imas_tools/input/core_profiles.py
+++ b/torax/_src/imas_tools/input/core_profiles.py
@@ -194,13 +194,7 @@ def plasma_composition_from_IMAS(
   impurity_species = {}
   main_ion_density = {}
   for ion in range(len(profiles_1d[0].ion)):
-    try:
-      symbol = str(profiles_1d[0].ion[ion].name)
-    except AttributeError:
-      # TODO(b/459479939): i/539) - Indicate supported dd_versions and switch on
-      # that instead of using a try-except.
-      # Case ids is plasma_profiles in early DDv4 releases.
-      symbol = str(profiles_1d[0].ion[ion].label)
+    symbol = str(profiles_1d[0].ion[ion].name)
     if symbol in parsed_ions:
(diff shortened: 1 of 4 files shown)


Nush395:
Does DD version 4.1.1 match to a specific version of imas python?

MateoBell:
No I think both are compatible with imas-python 2.1 or newer. It just needs specific version of this repo: https://github.com/iterorganization/IMAS-Data-Dictionaries, which is a dependency of imas-python.

Nush395:
> No I think both are compatible with imas-python 2.1 or newer. It just needs specific version of this repo: https://github.com/iterorganization/IMAS-Data-Dictionaries, which is a dependency of imas-python.

I see, so how would we verify that a TORAX user is using DD 4.1.1 ?

MateoBell:
Ah good point. I think we cannot know for sure. If it installs the dependencies it will automatically have 4.1.1 as default version and convert IDS to this version if needed (default in the ` loader.py` function). 
But for core_profiles or equilibrium inputs for example if it's 4/0 or 4.1.1 it would work in both cases as the IDS structures did not change (at least the field TORAX uses). And if it fails it is because a field has been renamed, but the conventions are the same between the two versions. 
Do you want TORAX to fail explicitly if the IDSs are not in the good version ?

MateoBell:
Also, we could add `imas-data-dictionaries` to TORAX dependencies to ensure the latest DD version (or specifically 4.1.1) is installed for TORAX users ?

Nush395:
> Also, we could add `imas-data-dictionaries` to TORAX dependencies to ensure the latest DD version (or specifically 4.1.1) is installed for TORAX users ?

That feels like a good solution!

PR #2203, opened 2026-06-09
Add normalized collisionality nu_star to outputs

r1cksync:
## What

Adds the electron–ion normalized collisionality `nu_star` to `PostProcessedOutputs`, exposing it as a face-grid simulation output. Resolves #2094.

`nu_star` is already computed by `collisions.calc_nu_star` (and used internally by the QuaLiKiz-based transport model). This change plumbs it through `PostProcessedOutputs` following the existing pattern (e.g. `FFprime`, `psi_norm`), so it flows automatically into the netCDF `profiles` group via `output.py` — no output-serialization changes needed.

A collisionality multiplier of `1.0` is used, so the reported value is the physical collisionality. (The multiplier is a QLKNN sensitivity-testing knob; happy to expose a configurable value instead if you'd prefer.)

## Changes
- `post_processing.py`: new `nu_star` field + docstring, `zeros()` initializer entry, and computation in `make_post_processed_outputs`.
- `collisions_test.py`: unit tests for `calc_nu_star` (previously untested) — shape, positivity/finiteness, and linear scaling with the collisionality multiplier.
- `post_processing_test.py`: asserts `nu_star` is exposed on the face grid.

## Testing
Ran locally with `TORAX_JAXTYPING=True TORAX_ERRORS_ENABLED=True`:
- `collisions_test.py` — 17/17 pass
- `post_processing_test.py::PostProcessingTest` — pass (incl. new `test_nu_star_output`)
- `output_test.py` profile/scalar shape tests — pass (confirms `nu_star` serializes correctly)

### DIFF
--- a/torax/_src/output_tools/tests/post_processing_test.py
+++ b/torax/_src/output_tools/tests/post_processing_test.py
@@ -326,6 +326,39 @@ class PostProcessingTest(parameterized.TestCase):
         rtol=1e-5,
     )
 
+  def test_nu_star_output(self):
+    """Checks nu_star is exposed as a physical, face-grid output."""
+    input_state = sim_state.SimState(
+        t=jnp.array(0.0),
+        dt=jnp.array(1e-3),
+        core_profiles=self.core_profiles,
+        core_transport=state.CoreTransport.zeros(self.geo),
+        core_sources=self.source_profiles,
+        geometry=self.geo,
+        solver_numeric_outputs=state.SolverNumericOutputs(
+            solver_error_state=np.array(0, jax_utils.get_int_dtype()),
+            outer_solver_iterations=np.array(0, jax_utils.get_int_dtype()),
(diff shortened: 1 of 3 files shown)


torax/_src/output_tools/tests/post_processing_test.py
```diff
@@ -326,6 +326,39 @@ def test_current_outputs(self):
         rtol=1e-5,
     )
 
+  def test_nu_star_output(self):
```
jcitrin, line 329: this test can be removed. It's redundant with the new unit tests in `collisions_test`

jcitrin (approved):
Sorry for the delayed review! This LGTM. Preemptively approving it although please remove the redundant test in output_tools.

Also, please resync with head and solve the merge conflict, then we can bring this in.

Note that in principle it's necessary to regenerate the test nc files since they'll have the new post processed outputs. However we can do that on our end, so no action needed.

PR #2214, opened 2026-06-15
Fix sign mismatch for psi and derived quantities

MateoBell:
Found a COCOS inconsistency for the sign of psi and related quantities such as dV/dpsi between IMAS using COCOS 17 and TORAX internal conventions. 
Mismatch had no consequence in the input mapping because of signs cancelling when building the intermediates, and also the reference IMAS equilibrium seemed to be not correctly written in COCOS 17 due to a bug in IMAS-python library used to convert it from DDv3.42 to DD4.0.0. Therefore also regenerated the equilibrium with flipped signs for the affected quantities.


torax/data/third_party/imas/ITERhybrid_COCOS17_IDS_ddv4.nc
Nush395, line 1: why did the IDS file change?

MateoBell:
After verification the signs seems to be consistent with COCOS 17.

PR #2243, opened 2026-06-23
feat: add q-profile warnings for unphysical configs (Issue #844)

### LINKED ISSUE #844: Provide helpful warnings when an unphysical config has been launched
Example issues

- [ ] When q-profile is clearly too low (due to too high Ip and low B)
- [ ] When the power / particle source densities are unphysically high

Paramveersingh-S:
Fixes #844.

This PR adds a safety check inside the `run_simulation` orchestration layer. If a user provides an initial `q` profile configuration where $q < 0.5$, it now immediately fires an `absl.logging.warning` to warn the user about potential unphysical simulation behavior (like premature crashes in tearing mode or sawtooth crash calculations) rather than letting it silently fail later in the solver. 

A unit test `test_q_profile_warning_logged` has been added to `run_simulation_test.py` to ensure this warning fires correctly.

### DIFF
--- a/torax/_src/orchestration/run_simulation.py
+++ b/torax/_src/orchestration/run_simulation.py
@@ -32,6 +32,8 @@ from torax._src.orchestration import sim_state
 from torax._src.orchestration import step_function
 from torax._src.output_tools import output
 from torax._src.output_tools import post_processing
+from absl import logging
+import numpy as np
 from torax._src.torax_pydantic import model_config
 import xarray as xr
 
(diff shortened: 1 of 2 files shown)


jcitrin:
Thanks for your desire to contribute to TORAX. Closing this PR since a more rigorous design is necessary. This kind of check on initial states should be an extensible method called within `get_initial_state_and_post_processed_outputs` , and done as a Jax callback so that the logging would actually work under JIT. This is something that we can do since we'd then like to design more extensive checks.

PR #2285, opened 2026-07-14
Reduce plotting slider payload for faster figure load

Neal006:
Fixes part of the slider payload bloat described in #553.

## Problem

`_get_slider_steps()` re-sent the full (static) x-array for every spatial
trace on every slider step, and encoded the timestamp indicator as a
per-step trace update instead of a lightweight shape move. Neither the
x-grid nor the vline's y-extent changes between steps — only trace y-values
and the vline's x-position do.

## Fix

- Timestamp indicator is now a `layout.shape` (moved via `relayout`) instead
  of a full trace re-sent on every step.
- Slider steps use Plotly's `update` method with separate restyle (`y` only)
  and relayout (`shapes[i].x0/x1`) payloads — `x` arrays are no longer
  duplicated per step.
- `updatemenus` mode-switch buttons now merge the relayout keys into their
  existing step-0 args instead of duplicating them.

## Measured (rampup reference dataset, min of 5 runs)

| Metric | Before | After | Δ |
|---|---|---|---|
| Figure JSON size | 1.81 MB | 1.31 MB | −28% |
| `to_json()` time | 0.089s | 0.059s | −34% |
| Figure build time | 0.851s | 0.824s | −3% (build is dominated by `.nc` load, not slider assembly — see note below) |

Build time is largely unaffected because it's dominated by data loading and
trace construction, not slider-step serialization. The measured win is in
payload size and serialize time, which is what actually affects load and
parse time client-side (browser fetch/parse/hold of the HTML figure).

This is one isolated speedup, per your request. A second, larger win is
still on the table: the `updatemenus` buttons currently embed a *full copy*
of both slider-mode step lists (linear time vs. simulation steps), so the
step data is stored ~2x beyond what's in this PR. That's a design choice
(which mode's steps live in the button vs. get rebuilt) I'd like input on
before touching it — happy to open a follow-up issue/PR once we agree on
approach.

## Tests

Added `SliderPayloadTest` in `plotruns_lib_test.py` (4 cases): steps carry
no `x` payload, timestamp indicator moves via shape relayout, timestamp
indicator is a shape not a trace, mode buttons still switch slider steps.
Full plotting suite: 236/236 passing.

### DIFF
--- a/torax/_src/plotting/plotruns_lib.py
+++ b/torax/_src/plotting/plotruns_lib.py
@@ -667,7 +667,6 @@ def _add_traces_and_update_axes(
               'trace_idx': trace_count,
               'attr': attr,
               'dataset': dataset,
-              'x': x,
           })
         else:
           # Time series plots (static)
@@ -693,26 +692,24 @@ def _add_traces_and_update_axes(
         )
         trace_count += 1
 
-    # Add vertical line to denote current slider time for time series plots
+    # Add vertical line to denote current slider time for time series plots.
+    # A layout shape rather than a trace, so slider steps can move it via a
+    # small relayout instead of re-sending x/y arrays per step.
     if not is_spatial:
       ylow, yhigh = _get_y_limits(datasets, axis_config)
-      fig.add_trace(
-          go.Scatter(
-              x=[datasets[0].t[0], datasets[0].t[0]],
-              y=[ylow, yhigh],
-              mode='lines',
-              name='Current Time',
-              showlegend=False,
-              line=dict(color='gray', width=1, dash='dot'),
-          ),
+      fig.add_shape(
+          type='line',
+          x0=datasets[0].t[0],
+          x1=datasets[0].t[0],
+          y0=ylow,
+          y1=yhigh,
+          line=dict(color='gray', width=1, dash='dot'),
           row=row,
           col=col,
       )
       timestamp_line_info.append({
-          'trace_idx': trace_count,
-          'y': [ylow, yhigh],
+          'shape_idx': len(timestamp_line_info),
       })
-      trace_count += 1
 
     # Update Axes Labels
     x_axis_kwargs = dict(
@@ -766,7 +763,6 @@ def _get_slider_steps(
 
   for t_val in t_vals:
     y_updates = []
-    x_updates = []
     trace_indices = []
 
     for info in spatial_traces_info:
@@ -775,18 +771,20 @@ def _get_slider_steps(
       nearest_t_idx = int(np.argmin(np.abs(dataset_t - t_val)))
       val_array = getattr(info['dataset'], info['attr'])
       y_updates.append(val_array[nearest_t_idx, :])
-      x_updates.append(info['x'])
       trace_indices.append(info['trace_idx'])
 
+    # The x-arrays of spatial traces are static rho grids, so only y is
+    # restyled. Timestamp vlines are layout shapes, moved via the relayout
+    # part of the 'update' method. This halves the per-step payload.
+    relayout_updates = {}
     for info in timestamp_line_info:
-      y_updates.append(info['y'])
-      x_updates.append([t_val, t_val])
-      trace_indices.append(info['trace_idx'])
+      relayout_updates[f'shapes[{info["shape_idx"]}].x0'] = t_val
+      relayout_updates[f'shapes[{info["shape_idx"]}].x1'] = t_val
 
     step = {
-        'method': 'restyle',
+        'method': 'update',
         'label': f'{t_val:.3f}s',
-        'args': [{'y': y_updates, 'x': x_updates}, trace_indices],
+        'args': [{'y': y_updates}, relayout_updates, trace_indices],
     }
     steps.append(step)
   return steps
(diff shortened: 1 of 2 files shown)


Neal006:
@jcitrin  , just checking up on updates regarding the PR approval. If there are any changes that you need me to do, I'm happy to do it.

Nush395:
Hi Neal, thanks for the change! On testing this out manually it does not noticeably improve the timings. 

To clarify understanding on the benchmark you did, where is the `to_json` in the timing table called in TORAX? This way I can understand the timing improvement as well. Thanks!

Neal006:
@Nush395 , 
Thanks for testing it. You are right, and my table overstated the effect.
`to_json` is never called in TORAX. The real path is `fig.show()` at
`plotruns_lib.py:497`, which goes to Plotly's `BrowserRenderer.render` and calls
`plotly.io.to_html(..., include_plotlyjs=True, validate=False)`. I benchmarked
`to_json()` standalone as a proxy for that step and left validation on, which
inflated the number I reported.
Re-measured on the real `show()` path with `test_iterhybrid_rampup.nc`, this
branch against `0d9546f3`:

| | before | after |
|---|---|---|
| figure data | 1.90 MB | 1.37 MB |
| plotly.js inlined | 4.84 MB | 4.84 MB |
| full HTML to browser | 6.74 MB | 6.22 MB |
| serialize | 0.114 s | 0.081 s |
| load + build + serialize | 0.963 s | 0.898 s |

The change saves 65 ms and 0.5 MB of a 6.2 MB page. plotly.js accounts for
4.84 MB of that page and figure build for 0.82 s of the 0.90 s, and this PR
touches neither. That is why manual testing shows no difference.

The dedup is still correct and the payload is smaller, but it is not a
user-visible speedup on its own. I am happy to close this if the churn is not
worth it, or keep it as a payload reduction with the description corrected.
Your call.

Nush395:
Maybe to take a step back and to understand the profile of why the slider is slow to use? You did the following profile in https://github.com/google-deepmind/torax/issues/553. 

| run | saved steps | figure JSON | redundant share |
|---|---|---|---|
| `basic_config` (t_final=5) | 22 | 0.9 MB | 55% |
| `basic_config` (t_final=25) | 106 | 4.2 MB | 56% |
| `iterhybrid_rampup` | 41 | 1.8 MB | 53% |

Can you help me to understand this profile as I don't fully follow.

Neal006:
Good question — that table was measuring the wrong thing for this issue, so let me
explain what it is and then give you the profile that actually answers "why is the
slider slow to use".

## What the #553 table measured

* **"figure JSON"** = `len(fig.to_json())`, i.e. the figure payload that gets embedded
  in the HTML page that `fig.show()` opens.
* **"redundant share"** = the fraction of those bytes that are duplicated slider data:
  (a) the static rho grids re-sent as `x` in every slider step, and (b) both slider-mode
  step lists embedded a second time inside the `updatemenus` button args, so per-timestep
  data is serialized roughly 3x (once in `layout.sliders`, twice in `layout.updatemenus`).

So it is a **page-weight** measurement. It says nothing about how long one slider move
takes, and I should not have presented it as a responsiveness profile. That is why your
manual test showed no difference.

## Profile of the actual interaction

Headless Chromium (Playwright), default plot config, `test_iterhybrid_rampup.nc`, each
step applied exactly as the slider applies it, timed to the second painted frame.
Medians of 3 runs of a full sweep over all steps.

**Per slider move — this branch changes nothing, as you found:**

| figure | subplots / traces | main | this PR |
|---|---|---|---|
| `iterhybrid_rampup` (41 steps) | 15 / 32 | 117 ms | 117 ms |
| `basic_config` t_final=25 (106 steps) | 15 / 33 | 133 ms | 119–133 ms |

Driving it through real mouse events on the slider rail instead of the API gives the
same range (~110–200 ms per move).

**Where those ~120 ms go (main, `iterhybrid_rampup`):**

| operation | ms |
|---|---|
| restyle all 32 traces (one slider step) | 98 |
| restyle **one** trace | 117 |
| `Plotly.redraw()` — no data change at all | 117 |
| empty restyle / relayout (measurement floor: 2 frames) | 33 |

Restyling one trace costs the same as restyling all of them, and a redraw that changes
no data at all costs the same again. **The cost is one full re-render of the figure, not
the data being pushed into it.** Hence payload size cannot move it.

It scales with how much figure there is to draw:

| plot config | subplots / traces | per move |
|---|---|---|
| `default_plot_config` | 15 / 32 | 98–117 ms |
| `simple_plot_config` | 6 / 12 | 52–67 ms |

I also checked one obvious suspect: pinning all axes to fixed ranges (no autorange)
changes nothing (only 3 of 30 axes autorange anyway).

## What does make it faster

Plotly's own slider pattern — `frames` + `method: 'animate'` with
`frame: {redraw: false}` — updates traces in place instead of re-rendering the figure:

| slider step mechanism | per move |
|---|---|
| current `restyle` / `update` | 88–118 ms |
| `frames` + `animate`, `redraw: false` | **50 ms** |

I prototyped this in the browser against the built figure and screenshotted both paths
at the same timestep — output is visually identical. Two things that have to be right:
each frame must carry **every** trace (a frame that omits the static time-history traces
blanks those panels), and the timestamp vlines ride along in `frame.layout`.

## Where that leaves this PR

This PR only reduces bytes: 1.90 MB -> 1.37 MB of figure data, 6.74 MB -> 6.22 MB of
page (4.84 MB of which is inlined plotly.js), 65 ms off a ~0.9 s build+serialize. It is
not a responsiveness fix and I should not have framed it as one. Happy to either close
it, or keep it narrowly as a payload reduction with the description corrected — your
call. Either way, if the `frames`+`animate` direction looks right to you I would rather
put the effort there, since that is the one that moves the number in #553.

Nush395:
> Either way, if the frames+animate direction looks right to you I would rather
> put the effort there, since that is the one that moves the number in https://github.com/google-deepmind/torax/issues/553.

This does sound like the better direction to move in for the feature improvement we want. I'll close this PR and please feel free to open another with the improvement.

PR #2291, opened 2026-07-16
Minor plot improvements

theo-brown:
- Order T and chi plot traces as electron-then-ion, matching density. This makes colors match across the plots.
- Add a dropdown to select which comparison runs are shown in the plot.
- Make legend double-click toggle only the clicked trace, not the whole subplot group.


jcitrin (approved):
Tested and the dropdown menu is indeed a great feature. Bringing it in.

PR #2313, opened 2026-07-28
Add a pellet-aware time step calculator.

MateoSochard:
New time step calculator named pellet_aware_time_step_calculator, created for the coupling between TORAX and HPI2-NN, that aligns time steps with pellet injection (trigger) times and ablation windows, resolving each ablation window as a single step. Generic for every pellet source with trigger_times/frequency and ablation_time in their runtime parameters. Also Added documentation for this new time step calculator and the HPI2-NN pellet source.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -2472,6 +2472,12 @@ implemented in TORAX.
 ``S_total`` (**time-varying-scalar** [default = 2e22])
   Total particle source in units of particles/s
 
+An external machine-learning surrogate of pellet ablation and deposition,
+HPI2-NN, can be registered as an alternative ``pellet`` model
+(``model_name: 'hpi2_nn'``) via ``torax.sources.register_source_model_config``.
+See the `HPI2-NN repository <https://github.com/DIFFER-NL/hpi2nn>`_ for
+installation and configuration.
+
 mhd
 ---
 
@@ -2692,7 +2698,7 @@ time_step_calculator
 
 ``calculator_type`` (str [default = 'chi'])
   The name of the ``time_step_calculator``, a method which calculates ``dt`` at
-  every timestep. Three methods are currently available:
+  every timestep. Four methods are currently available:
 
 * ``'fixed'``
     ``dt`` is equal to ``fixed_dt`` defined in :ref:`numerics_dataclass`.
@@ -2718,6 +2724,17 @@ time_step_calculator
      to a smaller value, rather than starting from max_dt every time. It can
      also speed up simulations where a small dt is required for a short section,
      but then a larger dt is be appropriate for the remainder of the simulation.
+
+* ``'pellet_aware'``
+    Uses a base calculator (``'chi'`` or ``'fixed'``) and additionally aligns
+    time steps with pellet trigger times and ablation windows, so that a step
+    lands exactly on each pellet and the ablation window is resolved as a single
+    step. It is generic over pellet sources: it reads ``trigger_times`` /
+    ``frequency`` and ``ablation_time`` from the configured pellet source's
+    runtime parameters, and if the source exposes it, a model predicted ablation
+    window. Intended for use with the HPI2-NN pellet model (see the ``pellet``
+    source above and the `HPI2-NN repository
+    <https://github.com/DIFFER-NL/hpi2nn>`_).
 
 ``tolerance`` (float [default = 1e-7])
   The tolerance within the final time for which the simulation will be
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -2458,8 +2458,9 @@ Ohmic power.
 pellet
 ^^^^^^
 
-Time-dependent Gaussian pellet source. No first-principle-based model is yet
-implemented in TORAX.
+Time-dependent Gaussian pellet source by default. A model of pellet ablation
+and deposition can also be registered as an alternative ``pellet`` model (see
+below).
 
 ``mode`` (str [default = 'model'])
 
@@ -2726,19 +2727,46 @@ time_step_calculator
      but then a larger dt is be appropriate for the remainder of the simulation.
 
 * ``'pellet_aware'``
-    Uses a base calculator (``'chi'`` or ``'fixed'``) and additionally aligns
-    time steps with pellet trigger times and ablation windows, so that a step
-    lands exactly on each pellet and the ablation window is resolved as a single
-    step. It is generic over pellet sources: it reads ``trigger_times`` /
-    ``frequency`` and ``ablation_time`` from the configured pellet source's
-    runtime parameters, and if the source exposes it, a model predicted ablation
-    window. Intended for use with the HPI2-NN pellet model (see the ``pellet``
-    source above and the `HPI2-NN repository
+    Uses a configurable base calculator (see ``base_calculator`` below) and
+    additionally aligns time steps with pellet trigger times and ablation
+    windows, so that a step lands exactly on each pellet and the ablation window
+    is resolved as a single step. It is generic over pellet sources: it reads
+    ``trigger_times`` / ``frequency`` and ``ablation_time`` from the configured 
+    pellet source's runtime parameters and if the source exposes it, a model 
+    predicted ablation window. Intended for use with the HPI2-NN pellet model 
+    (see the ``pellet`` source above and the `HPI2-NN repository
     <https://github.com/DIFFER-NL/hpi2nn>`_).
+
+    In the ``frequency`` mode the pellet source may toggle injection on and off
+    over time. Because floating point means a step never lands exactly on a
+    period boundary, such a toggle should switch on a small margin before the
+    intended pellet time, otherwise the pellet at that boundary may be skipped.
 
 ``tolerance`` (float [default = 1e-7])
   The tolerance within the final time for which the simulation will be
   considered done.
+
+The following attributes only apply to the ``'pellet_aware'`` calculator:
+
+``base_calculator`` (dict [default = None])
+  Configuration of the base time step calculator used away from pellet events,
+  for example ``{'calculator_type': 'chi'}`` or ``{'calculator_type':
+  'fixed'}``. It cannot itself be ``'pellet_aware'``. If None, the ``'chi'``
+  calculator is used.
+
+``trigger_tolerance`` (float [default = 1e-8])
+  Time tolerance for deciding whether the current time is at a pellet trigger.
+  The pellet source's own ``trigger_tolerance`` is used instead when it exposes
+  one, so that the step alignment and the source's deposition agree on when a
+  pellet fires.
+
+``window_after_pellet`` (float [default = 0.0])
+  Duration of the window after a pellet trigger during which ``dt_after_pellet``
+  is used. Not used by default.
+
+``dt_after_pellet`` (float [default = None])
+  Time step used during the window after a pellet trigger. If None, the base
+  calculator's step is used. Not used by default.
 
 neoclassical
 ------------
--- a/torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
+++ b/torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
@@ -128,8 +128,12 @@ class PelletAwareTimeStepCalculator(time_step_calculator.TimeStepCalculator):
       # frequency is not None here (guaranteed by the check above).
       assert frequency is not None
       frequency_t_start = getattr(pellet_params, 'frequency_t_start', 0.0)
+      # Whether the periodic injector is on at t. Defaults to on, so a source
+      # that does not expose it keeps firing at every period.
+      injection_enabled = getattr(pellet_params, 'injection_enabled', True)
       dt_trigger, dt_after_trigger, at_trigger = self._dt_for_frequency(
-          t, dt_standard, frequency, frequency_t_start, ablation_window, tol
+          t, dt_standard, frequency, frequency_t_start, injection_enabled,
+          ablation_window, tol,
       )
 
     dt = jnp.minimum(dt_standard, jnp.minimum(dt_trigger, dt_after_trigger))
@@ -190,10 +194,15 @@ class PelletAwareTimeStepCalculator(time_step_calculator.TimeStepCalculator):
       dt_standard: jax.Array,
       frequency: jax.Array,
       frequency_t_start: float | jax.Array,
+      injection_enabled: bool | jax.Array,
       ablation_window: jax.Array,
       tol: jax.Array,
   ) -> tuple[jax.Array, jax.Array, jax.Array]:
     """Aligns dt with a periodic pellet injection frequency.
+
+    The frequency is assumed strictly positive (validated by the pellet source
+    config). injection_enabled toggles the injector on or off. When off, no
+    pellet fires and the base time step is used.
 
     Returns:
       (dt_trigger, dt_after_trigger, at_trigger).
@@ -202,30 +211,31 @@ class PelletAwareTimeStepCalculator(time_step_calculator.TimeStepCalculator):
     inf = jnp.asarray(jnp.inf, dtype=dtype)
     frequency = jnp.asarray(frequency, dtype=dtype)
     frequency_t_start = jnp.asarray(frequency_t_start, dtype=dtype)
-    positive_frequency = frequency > 0.0
-    safe_frequency = jnp.where(
-        positive_frequency, frequency, jnp.asarray(1.0, dtype=dtype)
-    )
-    period = 1.0 / safe_frequency
+    injection_enabled = jnp.asarray(injection_enabled, dtype=bool)
+    period = 1.0 / frequency
     # Phase measured from frequency_t_start (consistent with the source).
-    phase = jnp.mod(t - frequency_t_start + tol, period)
+    phase = jnp.mod(t - frequency_t_start, period)
     # Float rounding can leave phase just below period instead of wrapping
     # to 0 at a pellet time.
     phase = jnp.where(period - phase < tol, jnp.asarray(0.0, dtype=dtype), phase)
     delta_to_next_period = period - phase
-    active = jnp.logical_and(positive_frequency, t > frequency_t_start + tol)
     # A single step covers the whole ablation window, so we only need to detect
     # the firing instant (phase wrapped to 0, the same test the source should use).
-    at_trigger = jnp.logical_and(active, phase <= tol)
+    # A pellet fires when the step lands within tol of a period boundary.
+    # Floating point means the step never lands exactly on the boundary, so a
+    # time varying injection_enabled that switches on exactly at a pellet's time
+    # can still read "off" at the firing instant. Turn it on a small margin
+    # before the intended pellet time.
+    at_trigger = jnp.logical_and(injection_enabled, phase <= tol)
     dt_trigger_value = jnp.where(at_trigger, ablation_window, delta_to_next_period)
-    dt_trigger = jnp.where(active, dt_trigger_value, inf)
+    dt_trigger = jnp.where(injection_enabled, dt_trigger_value, inf)
 
     dt_after_trigger = dt_standard
     if self._dt_after_pellet is not None:
       dt_after_pellet = jnp.asarray(self._dt_after_pellet, dtype=dtype)
       window_after_pellet = jnp.asarray(self._window_after_pellet, dtype=dtype)
       in_post_pellet = jnp.logical_and(
-          active,
+          injection_enabled,
           jnp.logical_and(
               jnp.logical_and(phase > tol, jnp.logical_not(at_trigger)),
               phase < window_after_pellet - tol,
--- a/torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
+++ b/torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
@@ -1,0 +1,297 @@
+# Copyright 2024 DeepMind Technologies Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Time step calculator that aligns steps with pellet trigger windows."""
+
+import jax
+from jax import numpy as jnp
+from torax._src.config import runtime_params as runtime_params_lib
+from torax._src.orchestration import sim_state as sim_state_lib
+from torax._src.time_step_calculator import chi_time_step_calculator
+from torax._src.time_step_calculator import fixed_time_step_calculator
+from torax._src.time_step_calculator import time_step_calculator
+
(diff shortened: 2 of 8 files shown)


torax/_src/time_step_calculator/pydantic_model.py
```diff
@@ -32,7 +33,14 @@ class TimeStepCalculatorType(enum.Enum):
   CHI = 'chi'
   FIXED = 'fixed'
   FROM_PREVIOUS_DT = 'from_previous_dt'
+  PELLET_AWARE = 'pellet_aware'
 
+@enum.unique
+class BaseTimeStepCalculatorType(enum.Enum):
```
jcitrin: This should be a StrEnum

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Returns:
+      dt: Scalar time step duration.  
+  """
+
+  def __init__(
+      self,
+      base_calculator_type: str = 'chi',
+      trigger_tolerance: float = 1e-8,
+      pellet_source_name: str = 'pellet',
```
jcitrin: should not be needed. There is only ever a single pellet source name. Can hard-code it internally to 'pellet'

torax/_src/time_step_calculator/pydantic_model.py
```diff
@@ -41,12 +49,21 @@ class TimeStepCalculator(torax_pydantic.BaseModelFrozen):
     calculator_type: The type of time step calculator to use.
     tolerance: The tolerance within the final time for which the simulation will
       be considered done.
+    pellet_source_name: For the 'pellet_aware' calculator, the name of the pellet
+      source to align time steps with. Defaults to 'pellet'.
```
jcitrin: should not be needed. There is only ever a single pellet source name. Can hard-code it internally to 'pellet'

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+      runtime_params: runtime_params_lib.RuntimeParams,
+      sim_state: sim_state_lib.SimState,
+  ) -> jax.Array:
+    """Returns a dt aligned with pellet trigger and ablation windows."""
+    dt_standard = self._base_calculator._next_dt(runtime_params, sim_state)
+    dt_standard = jnp.asarray(dt_standard)
+    dtype = dt_standard.dtype
+
+    t = sim_state.t
+    pellet_params = runtime_params.sources.get(self._pellet_source_name)
```
jcitrin: just `pellet_params = runtime_params.sources.get('pellet')` and remove pellet_source_name from everywhere

torax/_src/time_step_calculator/pydantic_model.py
```diff
@@ -32,7 +33,14 @@ class TimeStepCalculatorType(enum.Enum):
   CHI = 'chi'
   FIXED = 'fixed'
   FROM_PREVIOUS_DT = 'from_previous_dt'
+  PELLET_AWARE = 'pellet_aware'
```
jcitrin, line 38: There is currently no validation that if the TimeStepCalculator is `PELLET_AWARE` , that an appropriate pellet model is actually being used.

Recommend making a new Pydantic validator on ToraxConfig . Here is an agent recommended function for that

```
@pydantic.model_validator(mode='after')
def _check_pellet_aware_time_step_calculator_compatibility(self) -> typing_extensions.Self:
  if self.time_step_calculator.calculator_type == time_step_calculator_pydantic_model.TimeStepCalculatorType.PELLET_AWARE:
    if self.sources.pellet is None:
      raise ValueError(
          "time_step_calculator.calculator_type='pellet_aware' requires a pellet "
          "source to be configured under sources.pellet."
      )
    has_trigger_times = hasattr(self.sources.pellet, 'trigger_times')
    has_frequency = hasattr(self.sources.pellet, 'frequency')
    has_ablation_time = hasattr(self.sources.pellet, 'ablation_time')
    if not ((has_trigger_times or has_frequency) and has_ablation_time):
      raise ValueError(
          f"time_step_calculator.calculator_type='pellet_aware' requires a pellet "
          f"source supporting discrete pellet injection ('trigger_times' or 'frequency', "
          f"and 'ablation_time'), e.g. 'hpi2_nn'. "
          f"Got pellet model_name='{self.sources.pellet.model_name}'."
      )
  return self
```

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+    if pellet_params is None:
+      return dt_standard
```
jcitrin: can remove this, it should not occur after the Pydantic validation that an appropriate pellet model is being used.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+
+    trigger_times = getattr(pellet_params, 'trigger_times', None)
+    frequency = getattr(pellet_params, 'frequency', None)
+    ablation_time = getattr(pellet_params, 'ablation_time', None)
+    if ablation_time is None:
+        return dt_standard
```
jcitrin: remove. Pydantic should take care of ensuring that this exists with some default

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+    # method returning (at_trigger, ablation_time).
+    ablation_step_fn = getattr(pellet_params, 'ablation_step', None)
+    use_model_ablation = bool(
+        getattr(pellet_params, 'use_model_ablation_time', False)
+    ) and callable(ablation_step_fn)
+    if use_model_ablation:
+      at_trigger, model_ablation_time = ablation_step_fn(
+          sim_state.geometry, sim_state.core_profiles
+      )
+      at_trigger = jnp.asarray(at_trigger)
```
jcitrin: why is this casting necessary? (also in other places). Shouldn't at_trigger already be the correct type?

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+      model_ablation_time = jnp.asarray(model_ablation_time, dtype=dtype)
+
+    if trigger_times is not None:
```
jcitrin, line 123: The branching is quite confusing.

Recommend having two functions: one for the fixed trigger times and the other for frequency. That will at least avoid this big outer if else branching.

Before checking which function to call (based on existence if trigger_times or frequency), also good to have an assertion that one and only one of them is not None

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+          is_next_trigger,
+          jnp.minimum(next_trigger, trigger),
+          next_trigger,
+        )
+        is_past = trigger <= t + tol
+        last_trigger = jnp.where(
+          is_past,
+          jnp.maximum(last_trigger, trigger),
+          last_trigger,
+        )
```
jcitrin: Is a loop really needed for this? Can vectorize, like

```
triggers = jnp.asarray(trigger_times, dtype=dtype)
# Next trigger: earliest trigger > t - tol (fallback to inf)
next_trigger = jnp.min(jnp.where(triggers > t - tol, triggers, jnp.inf))
# Last trigger: most recent trigger <= t + tol (fallback to -inf)
last_trigger = jnp.max(jnp.where(triggers <= t + tol, triggers, -jnp.inf))
```

and then also no need to initialize last_trigger and next_trigger

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+    ablation_remaining = inf
+    dt_trigger = inf
+    next_trigger = inf
+    dt_after_trigger = inf
+    at_trigger = jnp.asarray(False)
+    model_ablation_time = inf
+
+    # A pellet source that predicts its own ablation time exposes a
+    # 'use_model_ablation_time' flag and an 'ablation_step(geo, core_profiles)'
+    # method returning (at_trigger, ablation_time).
```
jcitrin: This, together with `trigger_times`, `frequency`, `ablation_time` crystallizes a discrete pellet model API.

This is fine, but we'd then need to enforce it on the TORAX config level (e.g. with Pydantic validation) and mention this in the docs . This is however a separate PR so I'd make a TODO and we can think about it later.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+    ablation_step_fn = getattr(pellet_params, 'ablation_step', None)
+    use_model_ablation = bool(
+        getattr(pellet_params, 'use_model_ablation_time', False)
+    ) and callable(ablation_step_fn)
+    if use_model_ablation:
+      at_trigger, model_ablation_time = ablation_step_fn(
```
jcitrin: what does `at_trigger` represent? Comments here would be useful

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+          is_past,
+          jnp.maximum(last_trigger, trigger),
+          last_trigger,
+        )
+
+      delta_to_next_trigger = next_trigger - t
+      if use_model_ablation:
+        # One step at the trigger covering the model-predicted ablation time.
+        # ablation_remaining is only read when in_ablation.
+        in_ablation = at_trigger
```
jcitrin: in_ablation isn't used inside this block. Is it used elsewhere and needs to be defined here? Either way it's confusing :)

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+        ablation_remaining = model_ablation_time
+        dt_trigger = jnp.where(
+            at_trigger, model_ablation_time, delta_to_next_trigger
+        )
+      else:
+        end = next_trigger + ablation_time
+        in_ablation = jnp.logical_and(
+          t >= next_trigger - tol,
+          t < end - tol,
+        )
```
jcitrin: (agent found this one)

Is this a bug? If we're looking to see if `t` is sandwiched between a t_trigger and  t_trigger+t_ablation, then shouldn't we have

`in_ablation = jnp.logical_and(t >= last_trigger - tol, t < last_trigger + ablation_time - tol)`

(and remove the `end` line)

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+            jnp.logical_and(
+                jnp.logical_and(phase > tol, jnp.logical_not(in_ablation)),
+                phase < window_after_pellet - tol,
+            ),
+        )
+        post_remaining_freq = window_after_pellet - phase
+        dt_after_trigger = jnp.where(
+            in_post_pellet_freq,
+            jnp.minimum(dt_after_pellet, post_remaining_freq),
+            inf,
```
jcitrin: maybe just `dt_standard`?

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+                    t > last_trigger + tol, jnp.logical_not(in_ablation)
+                ),
+                t < post_window_end - tol,
+            ),
+        )
+        post_remaining = post_window_end - t
+        dt_after_trigger = jnp.where(
+            in_post_pellet,
+            jnp.minimum(dt_after_pellet, post_remaining),
+            inf,
```
jcitrin: maybe just `dt_standard`?

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+    elif base_calculator_type == 'fixed':
+      self._base_calculator = (
+          fixed_time_step_calculator.FixedTimeStepCalculator()
+      )
+    else:
+      raise ValueError(
+          'base_calculator must be "chi" or "fixed" '
+      )
+
+  def _next_dt(
```
jcitrin, line 74: There are no unit tests. Please add extensive unit tests of the various logic implemented here. Can use mocks such that only the specific runtime_params used in this function need to be set

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+        t + dt > runtime_params.numerics.t_final
+    )
+    dt = jax.lax.select(
+        jnp.logical_and(
+            runtime_params.numerics.exact_t_final,
+            crosses_t_final,
+        ),
+        runtime_params.numerics.t_final - t,
+        dt,
+    )
```
jcitrin: Is this needed here? The calling function does this I believe

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+    elif frequency is not None:
+      frequency = jnp.asarray(frequency, dtype=dtype)
+      frequency_t_start = jnp.asarray(
+          getattr(pellet_params, 'frequency_t_start', 0.0), dtype=dtype
+      )
+      positive_frequency = frequency > 0.0
+      safe_frequency = jnp.where(
+          positive_frequency, frequency, jnp.asarray(1.0, dtype=dtype)
+      )
+      period = 1.0 / safe_frequency
```
jcitrin: Positivity should be enforced on the Pydantic config level and assumed > 0.0 here.

What should definitely NOT be done is silently change the frequency from an invalid one to an arbitrary valid one

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+      # Phase measured from frequency_t_start (consistent with the source).
+      phase = jnp.mod(t - frequency_t_start + tol, period)
+      # Float rounding can leave phase just below period instead of wrapping
+      # to 0 at a pellet time.
+      phase = jnp.where(
+          period - phase < tol, jnp.asarray(0.0, dtype=dtype), phase
+      )
+      delta_to_next_period = period - phase
+
+      after_start = t > frequency_t_start + tol
```
jcitrin: this feels a bit limiting that we can only go from OFF to ON with pellets when using the frequency pathway.

There should be a `frequency_t_start` to measure the phase from some time.

But then there should also be another attribute in the pellet model runtime_params, a boolean time varying array (which is always constrained to be a STEP time-varying array), which toggles the pellet injector to be ON (true) or OFF (false), at the set frequency.

This runtime_param can be checked in the beginning of the branch (or new function) with frequency, and if it's false just return dt_standard immediately.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+      if use_model_ablation:
+        in_ablation = at_trigger
+        ablation_remaining = model_ablation_time
+        dt_trigger_value = jnp.where(
+            at_trigger, model_ablation_time, delta_to_next_period
+        )
+      else:
+        in_ablation = jnp.logical_and(
+            jnp.logical_and(positive_frequency, after_start),
```
jcitrin: this shouldn't be needed. We should validate positive_frequency on the Pydantic level, and as mentioned above, we should only be in this code pathway if the pellet injector is "ON" as seen by the new attribute.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+            ablation_time - phase,
+            ablation_remaining,
+        )
+        dt_trigger_value = jnp.where(
+            in_ablation,
+            ablation_remaining,
+            delta_to_next_period,
+        )
+      dt_trigger = jnp.where(
+          jnp.logical_and(positive_frequency, after_start),
```
jcitrin: as above

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+          dt_trigger_value,
+          dt_trigger,
+      )
+      if self._dt_after_pellet is not None:
+        dt_after_pellet = jnp.asarray(self._dt_after_pellet, dtype=dtype)
+        in_post_pellet_freq = jnp.logical_and(
+            jnp.logical_and(positive_frequency, after_start),
```
jcitrin: as above

docs/configuration.rst
jcitrin: This can be updated

docs/configuration.rst
```diff
@@ -2719,6 +2725,17 @@ time_step_calculator
      also speed up simulations where a small dt is required for a short section,
      but then a larger dt is be appropriate for the remainder of the simulation.
 
+* ``'pellet_aware'``
```
jcitrin, line 2729: list here and explain in more detail all the new attributes of the pellet_aware time_step_calculator

jcitrin (changes requested):

jcitrin:
@tamaranorman could you also take a look at reviewing this one?

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+from jax import numpy as jnp
+from torax._src.config import runtime_params as runtime_params_lib
+from torax._src.orchestration import sim_state as sim_state_lib
+from torax._src.time_step_calculator import chi_time_step_calculator
+from torax._src.time_step_calculator import fixed_time_step_calculator
+from torax._src.time_step_calculator import time_step_calculator
+
+
+class PelletAwareTimeStepCalculator(time_step_calculator.TimeStepCalculator):
+  """TimeStepCalculator that resolves pellet trigger and ablation windows.
```
tamaranorman, line 27: Is this something that could be made generic to many other discrete events to allow supporting other things moving forward

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
```diff
@@ -0,0 +1,297 @@
...
+
+  The pellet_aware time step calculator ensures that time steps are aligned with
+  pellet trigger times and ablation windows. 
+
+  It checks the current simulation time against the pellet trigger times 
+  and ablation duration, and adjusts the time step to ensure that steps 
+  do not skip over these events. 
+  
+  Arguments:
+    base_calculator_type: The type of the base time step calculator to use for
```
tamaranorman: Should this be instead base_calculator to make sure that any parameter are created if a calculator has these - then it can be something other than chi or fixed potentially

torax/_src/time_step_calculator/pydantic_model.py
MateoSochard: base_calculator is now a nested time step calculator config object instead of a string enum, so this enum is gone.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Done

torax/_src/time_step_calculator/pydantic_model.py
MateoSochard: Done

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Done

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Done

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Done

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: at_trigger is a boolean that is true when the current step lands on a pellet firing instant (within tol). The  logic was also reworked so the same at_trigger detection now drives both cases uniformly, whether the ablation window comes from the fixed ablation_time value or from the pellet model's predicted ablation time instead of having separate branches.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Instead of fixing the in_ablation window, I reworked the logic: the whole ablation is now resolved as a single step that lands on the trigger, so in_ablation and the end line are removed entirely and this bug can no longer occur. The calculator only detects the firing instant (at_trigger) and returns the ablation window as one step. It was done in order use have the same logic for the fixed ablation_time and the model's predicted ablation time.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Done

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: That's true. Frequency positivity is now validated in the pellet source config (this lives in the HPI2-NN model repo, so it isn't visible in this PR), and the calculator assumes frequency > 0. The silent clamping to a valid frequency is gone.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Added. The pellet source now exposes frequency_t_start and an injection_enabled flag (a STEP boolean time-varying param). These live in the pellet source config (HPI2-NN repo, not in this PR). The calculator reads them and returns dt_standard immediately when injection is off. Because a step never lands exactly on a period boundary, injection_enabled should switch on a small margin before the intended pellet time, this is documented in the calculator and in configuration.rst.

docs/configuration.rst
MateoSochard: Done

docs/configuration.rst
MateoSochard: Done

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: For this PR I have kept it only for pellets.

torax/_src/time_step_calculator/pellet_aware_time_step_calculator.py
MateoSochard: Done

jcitrin (approved):
Thanks for this feature! Any additional minor changes we can take care of internally since I am aware that you are at the end of your project.

PR #2314, opened 2026-07-28
Add crash suppression windows to sawtooth trigger models.

MateoSochard:
Add suppression_times and suppression_duration to the base sawtooth TriggerConfig, while the current time is within [t_event, t_event + duration) of any suppression_times entry, crashes are suppressed. Both default to empty/zero. This lets crashes be blocked around a prescribed event. It was created for the coupling between TORAX and HPI2-NN, where pellet injection can currently conflict with the simple sawtooth model. It provides a tool to handle that case if it happens. The tool itself is generic.

The logic lives in the base TriggerModel.__call__, which now delegates the trigger condition to a new abstract compute_trigger, trigger models(e.g. SimpleTrigger) define compute_trigger instead of __call__.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -2497,6 +2497,15 @@ sawtooth
   The minimum normalized radius (:math:`\hat{\rho}`) of the q=1 surface required
   to trigger a crash.
 
+* ``suppression_times`` (tuple[float, ...] [default = ()])
+  Times [s] of events after which sawtooth crashes are suppressed. Empty by default (no suppression). 
+  Available on all trigger models.
+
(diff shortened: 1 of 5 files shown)


jcitrin:
The necessity of this PR spurred a deeper investigation.

We found that indeed there was a bug in the sawtooth redistribution for hollow temperature profiles, leading to spurious negative temperatures.

This has been fixed in https://github.com/google-deepmind/torax/pull/2326

Thanks for helping expose this bug!

I am closing this PR now since it should be unnecessary. Please merge 2326 into your workflow and investigate if the cases that previously had negative temperatures during the sawtooth crash, now behave as expected. I'd be curious to hear any feedback on that.

PR #2318, opened 2026-07-30
Animate prebuilt frames from the plotting slider

### LINKED ISSUE #553: Speed up plotting slider responsiveness

Neal006:
# Animate prebuilt frames from the plotting slider

Fixes #553. Follow-up to #2285, which I closed after measuring that payload size
is not what makes the slider slow — this PR targets the thing that is.

## Why the slider is slow

From the profile in https://github.com/google-deepmind/torax/pull/2285#issuecomment-5113182237:
one slider move costs a full re-render of the figure. Restyling one trace costs
the same as restyling all 32, and a `Plotly.redraw()` that changes no data at all
costs the same again. Slider steps currently carry the data and apply it with
`Plotly.restyle`, so every move pays for a re-render.

## What this changes

* Each timepoint is stored once as a Plotly **frame** (`fig.frames`), holding `y`
  for the profile traces and `x` for the "current time" vertical lines. `x` for
  the profiles is static, so it stays on the trace and is no longer re-sent.
* Slider steps become `method: 'animate'` with `frame: {redraw: false}`: they
  reference a frame name instead of carrying data, and the traces are updated in
  place instead of the figure being re-rendered.
* Both slider modes ("Plasma time", "Simulation steps") reference the same
  frames, so per-timepoint data is no longer serialized once per mode inside
  `updatemenus`.

## Per-move latency

Headless Chromium (Playwright), `default_plot_config`, each step applied exactly
as plotly's slider applies it, timed to the second painted frame. Both pages are
open at once and sweeps are interleaved A/B — absolute numbers move by up to 2x
between runs on my machine, so a non-interleaved comparison is not trustworthy.
Median of 5 rounds of 25 moves.

| run | saved timesteps | main | this PR |
|---|---|---|---|
| `test_iterhybrid_rampup` | 41 | 153 ms | **67 ms** |
| `basic_config`, `t_final=25` | 106 | 176 ms | **67 ms** |

The new path does not scale with the number of saved timesteps; the old one does.
Dragging the rail with real mouse events shows the same behaviour.

## Payload and build

Medians of 3 repeats in-process (`create_plotly_figure` + `plotly.io.to_html`,
the `fig.show()` path).

| | rampup, main | rampup, PR | 106-step, main | 106-step, PR |
|---|---|---|---|---|
| figure JSON | 1.90 MB | 0.44 MB | 4.45 MB | 1.09 MB |
| full HTML page | 6.74 MB | 5.29 MB | 9.30 MB | 5.94 MB |
| figure build | 0.29 s | 0.37 s | 0.52 s | 0.45 s |
| serialize | 0.102 s | 0.061 s | 0.249 s | 0.129 s |

Build time goes up by ~0.08 s on the 41-step run: frames are validated by
plotly.py as (partial) traces. It goes down on the 106-step run because the
profiles are now fetched from the `PlotData` once per trace instead of once per
trace per slider step. Two slider moves pay for the worst case either way.

Time from opening the page to the first painted plot is unchanged (~1.1 s,
median of 5 fresh loads, both runs): that is dominated by parsing the 4.84 MB of
inlined plotly.js, which this PR does not touch. Consistent with the manual
testing on #2285 — page weight was never the bottleneck.

## Behaviour

Compared per-trace `x`/`y` of every trace between main and this branch, in both
slider modes, at 4 timepoints each:

* **Simulation steps** mode: identical, every trace, every probed step.
* **Plasma time** mode: all profiles identical. The only difference is the
  "current time" vertical lines, which now sit at the actual saved time of the
  profile being displayed instead of the interpolated tick value. Both modes
  share one frame per saved timestep, so the line follows the data rather than
  the tick. Worst-case offset is half a saved-timestep interval: 0 s for
  `test_iterhybrid_rampup` (uniform `dt`), 0.12 s of a 25 s axis for the
  adaptive-`dt` 106-step run. If you would rather keep the line exactly on the
  tick, say so and I will label the ticks with the frame's real time instead
  (one line), or give the linear-time mode its own frames for the line position.

Screenshots of the whole figure at the same timepoint are pixel-identical apart
from the slider handle, which now follows the animated frame (plotly syncs it
for `animate` steps) instead of staying where it was left.

## Tests

`SliderFramesTest` in `plotruns_lib_test.py` covers: steps animate to frames
that exist and carry no data; frames carry each timepoint's profile and no
static `x`; frames move the current-time lines; both modes share the frames and
reset to the first timepoint. Full plotting suite: 235 passed.

Also checked manually that comparing two runs with different time grids still
picks each dataset's nearest timepoint per frame.

## Reproducing

Build a page with `plotruns_lib.create_plotly_figure` + `plotly.io.to_html`,
then drive `layout.sliders[0].steps[i]` through `Plotly.animate` / `Plotly.restyle`
in Playwright, timing to the second `requestAnimationFrame`. Happy to attach the
scripts if useful.

### DIFF
--- a/torax/_src/plotting/plotruns_lib.py
+++ b/torax/_src/plotting/plotruns_lib.py
@@ -748,14 +748,102 @@ def _add_traces_and_update_axes(
   return spatial_traces_info, timestamp_line_info
 
 
-def _get_slider_steps(
+# Animation options used by the slider steps. 'immediate' jumps straight to the
+# requested frame, and redraw=False updates the traces in place instead of
+# re-rendering the whole figure, which dominates the cost of a slider move.
+_ANIMATION_ARGS: Final[dict[str, Any]] = {
+    'mode': 'immediate',
+    'frame': {'duration': 0, 'redraw': False},
+    'transition': {'duration': 0},
+}
+
+
+def _get_trace_update(
+    t_val: float,
+    spatial_traces_info: list[dict[str, Any]],
+    timestamp_line_info: list[dict[str, Any]],
+) -> tuple[dict[str, Any], list[int]]:
+  """Returns the (restyle args, trace indices) showing a single timepoint."""
+  y_updates = []
+  x_updates = []
+  trace_indices = []
+
+  for info in spatial_traces_info:
+    # Find the nearest time index in this dataset for the target time.
+    dataset_t = info['dataset'].t
+    nearest_t_idx = int(np.argmin(np.abs(dataset_t - t_val)))
+    val_array = getattr(info['dataset'], info['attr'])
+    y_updates.append(val_array[nearest_t_idx, :])
+    x_updates.append(info['x'])
+    trace_indices.append(info['trace_idx'])
+
+  for info in timestamp_line_info:
+    y_updates.append(info['y'])
+    x_updates.append([t_val, t_val])
+    trace_indices.append(info['trace_idx'])
+
+  return {'y': y_updates, 'x': x_updates}, trace_indices
+
+
+def _build_frames(
     data1: PlotData,
     spatial_traces_info: list[dict[str, Any]],
     timestamp_line_info: list[dict[str, Any]],
+) -> list[go.Frame]:
+  """Builds one animation frame per timestep of the master clock.
+
+  Slider steps animate to these frames rather than carrying the data
+  themselves, so that each timepoint is stored once and moving the slider
+  updates the traces in place instead of re-rendering the whole figure.
+
+  Args:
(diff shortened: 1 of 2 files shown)


Nush395 (approved):
This looks to speed up the slider. Thanks for the change!

Neal006:
@Nush395 , checking on the update of this PR, would like to know if there's anything I can do from my side.
Thanks!!

Nush395:
> @Nush395 , checking on the update of this PR, would like to know if there's anything I can do from my side. Thanks!!

Hi @Neal006, it's all good to bring in our side, that's my bad for being a bit slow. Will do later today.

PR #2383, opened 2026-08-19
Add TokaMaker geometry interface

fs2819:
This PR adds the ability to pass geometry information from [TokaMaker](https://github.com/openfusiontoolkit/openfusiontoolkit) directly (i.e. in memory) to TORAX via the 'tokamaker' geometry mode. The geometry information is calculated by TokaMaker (using the new `get_fsa()` function) and then passed to TORAX as a dict of arrays. 

This geometry mode bypasses the saving and reading of `eqdsk` files, which is especially useful for the TokaMaker+TORAX pulse design workflow (which currently requires saving and loading hundreds of `eqdsk` files). 

This does not add a dependency for TokaMaker (or OFT) to TORAX. The geometry inputs are simply floats and NumPy arrays. The new `tokamaker_test.py` added for this geometry type creates TokaMaker-like dict of arrays from a dummy cylindrical geometry, not requiring TokaMaker.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1141,6 +1141,15 @@ geometry
     (IDS) or directly the equilibrium IDS on the fly. It handles IDSs in Data
     Dictionary version 4.0.0.
 
+* ``'tokamaker'``
+    Takes flux surface averages computed by
+    `TokaMaker <https://github.com/openfusiontoolkit/OpenFUSIONToolkit>`_.
+    TokaMaker traces flux surfaces on the finite element mesh used for the
+    equilibrium solve, then passes arrays of flux surface averaged quantities
+    to TORAX. TORAX does not depend on TokaMaker (or Open FUSION Toolkit),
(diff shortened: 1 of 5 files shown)


fs2819:
@jcitrin @tamaranorman Let me know if you have any comments or suggestions for this PR. Thanks!

fs2819:
@jcitrin @tamaranorman As we discussed, I looked into passing an IMAS equilibrium object and memory and it seems possible but I think is overkill for the amount of information that needs to be passed. We have in our plan to make TokaMaker (and TokaMaker+TORAX) IMAS compatible in terms of inputs and outputs but not in the very near term. That said, I think the capability this PR adds is worthwhile to merge into TORAX. It's fast and lightweight, I have been running TMTX with it for weeks now. It's primarily new code so nothing that affects other use cases. I'm happy to be the main contact if other users want to use the TokaMaker interface, as well.

Let me know what you think.

PR #2400, opened 2026-08-25
Preserve Pereverzev transport under adaptive pedestal scaling

TheodoreWolf:
## Summary

- keep Pereverzev-Corrigan numerical transport coefficients outside adaptive pedestal clipping, scaling, and smoothing
- derive protected fields from the PereverzevTransport dataclass so all six paired terms remain consistent
- update the existing modifier test and add a JIT production-path regression comparing assembled coefficients with raw Pereverzev transport

## Context

Pereverzev diffusion and counter-convection terms are constructed as zero-net-flux pairs for numerical stabilization. Adaptive pedestal processing previously modified the Pereverzev chi, D, and particle V fields while leaving both full_v_heat counter-convection fields unchanged. Later clipping and pedestal-top smoothing amplified this mismatch.

## Validation

- 87 tests and 35 subtests passed with TORAX runtime assertions enabled
- focused pylint check: 10.00/10
- 60 s adaptive L-H simulation: 945 to 87 accepted steps; median accepted timestep 0.015625 s to 1.0 s; maximum normalized Pereverzev pair defect 0.973 to 1.9e-16

### DIFF
--- a/torax/_src/pedestal_model/pedestal_model_output.py
+++ b/torax/_src/pedestal_model/pedestal_model_output.py
@@ -25,8 +25,15 @@ from torax._src import state
 from torax._src.geometry import geometry
 from torax._src.internal_boundary_conditions import internal_boundary_conditions as internal_boundary_conditions_lib
 from torax._src.pedestal_model import runtime_params as pedestal_runtime_params_lib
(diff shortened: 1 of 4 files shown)


TheodoreWolf:
My agent got overeager: still working on this!

PR #2420, opened 2026-09-03
Fix Pereverzev transport coefficients modified under adaptive pedestal scaling

TheodoreWolf:
## Summary

I noticed when running simulations with [plasmax](https://github.com/TheodoreWolf/plasmax) many numerical errors when I turned `adaptive_transport` on. 
After some digging with agents, I found out that Pereverzev-Corrigan transport coefficients were scaled in the pedestal.

## Fix

- keep all Pereverzev-Corrigan coefficients unchanged under adaptive pedestal transport suppression so paired diffusion and counter-convection continue to cancel.
- add a test for this and fix the previous test.

## Numerical reproduction

I ran the existing ITER-hybrid adaptive L-H transition fixture (`torax/tests/test_data/test_iterhybrid_lh_transition.py`) to 60 s in three cases:

- `main`: existing behavior at `3ef7bdec`
- `main, P-C disabled`: the same revision with `solver.use_pereverzev=False`, as a diagnostic control
- `this PR`: the same configuration with the fix

| case | accepted steps | median dt [s] | retried steps | max P-C balance defect |
|---|---:|---:|---:|---:|
| main | 945 | 0.015625 | 883 | 0.973 |
| main, P-C disabled | 96 | 0.5 | 36 | NA |
| this PR | 87 | 1.0 | 33 | ~2e-16 |

All three runs reached 60 s with `NO_ERROR`.

### DIFF
--- a/torax/_src/pedestal_model/pedestal_model_output.py
+++ b/torax/_src/pedestal_model/pedestal_model_output.py
@@ -25,8 +25,15 @@ from torax._src import state
 from torax._src.geometry import geometry
 from torax._src.internal_boundary_conditions import internal_boundary_conditions as internal_boundary_conditions_lib
 from torax._src.pedestal_model import runtime_params as pedestal_runtime_params_lib
+from torax._src.transport_model import pereverzev as pereverzev_lib
 
 # pylint: disable=invalid-name
+
+
+_PEREVERZEV_FIELDS = frozenset(
+    field.name
+    for field in dataclasses.fields(pereverzev_lib.PereverzevTransport)
+)
 
 
 @jax.tree_util.register_dataclass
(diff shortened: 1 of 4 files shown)


jcitrin:
Hi Theo. Thanks a lot for this. We coincidentally noticed this ourselves this week. A fix for that as well as another bug in the Pereverzev coefficients as used in adaptive pedestal transport, are part of a PR chain currently being prepared.

We'll keep this PR open for now in case it isn't covered there, but in principle it will be a duplicate of some incoming internal work.

TheodoreWolf:
Thanks for the update! This PR actually follows up on #2400, which I opened on 25 August and closed while I finished validating the fix. Would you mind linking back to these PRs from the eventual fix? Hopefully the reproduction and regression test here are useful too.

PR #2422, opened 2026-09-04
Pre-interpolate time-dependent geometries for fixed time-step simulations

### LINKED ISSUE #2315: Speed improvement: pre-interpolate geometries when using `fixed_time_step`
`fixed_time_step` loop calls the geometry provider's time-interpolation once per step. As a result, loading multiple geometry time slices pays the interpolation cost on every step rather than once at the geometry initialisation, inflating both compile time (~2-5x) and run time (scaling with step count).

However, with a fixed dt, every future time point is knowable in advance, so we could in theory make `fixed_time_step` pre-interpolate all of the geometries.

wtfPrethiv:
Fixes #2315.

## Summary

For fixed time-step simulations with a constant `fixed_dt`, precompute time-dependent geometries once at setup and use a cheap nearest-time lookup during the simulation.

All other configurations preserve the existing interpolating-provider behavior.

## Validation

* `pyink` check passes for changed Python files
* `git diff --check` passes
* Targeted geometry, precompute-guard, time-grid, single-step-equivalence, and full-simulation-equivalence tests passed
* Broader orchestration and JIT suites were skipped locally because WSL became unresponsive during long JAX test runs

## Notes

The precomputed provider is publicly exported from `torax.experimental.geometry` for explicit opt-in use by the coupling path in a future change.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -2662,6 +2662,12 @@ time_step_calculator
     ``numerics`` config dict, then in practice some steps may have lower ``dt``
     if the solver needed to backtrack.
 
+    When all time steps are known in advance (constant ``fixed_dt``,
+    ``adaptive_dt==False``, no sawtooth model and no restart), a time-dependent
+    geometry is interpolated once onto every time step before the simulation
+    starts, instead of on every step. This reduces both compile time and run
+    time, at the cost of storing one geometry per time step in memory.
+
 * ``'chi'``
     adaptive dt method, where ``dt`` is a multiple of a base dt inspired by the
     explicit stability limit for parabolic PDEs:
--- a/torax/_src/geometry/geometry_provider.py
+++ b/torax/_src/geometry/geometry_provider.py
@@ -17,6 +17,7 @@
 NOTE: Time dependent providers currently live in `geometry.py` and match the
 protocol defined here.
 """
+
 from collections.abc import Mapping
 import dataclasses
 import functools
@@ -26,6 +27,7 @@ import chex
 import jax
 from jax import numpy as jnp
 import numpy as np
+from torax._src import array_typing
 from torax._src import interpolated_param
 from torax._src import jax_utils
 from torax._src.geometry import geometry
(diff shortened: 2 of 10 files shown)


sylvesterkaczmarek (commented):
The 10,000-geometry guard is applied too late to bound setup cost: `get_time_grid()` materialises every timestamp first, and only then does `_maybe_precompute_geometry_provider()` check `len(times)`. A tiny positive `fixed_dt` over a long interval can therefore allocate/iterate over millions of entries before deciding not to precompute. Could the count be checked analytically, or stop grid generation once the cap is exceeded?

wtfPrethiv:
Thanks for catching this. You’re right that checking the cap after generating the full grid doesn’t actually bound the setup work. I'll change the grid generation to stop once it goes over the precompute limit and fall back to the existing interpolating provider.

This approach seems safer than calculating the count upfront, since the grid follows the same floating-point accumulation and `exact_t_final` behavior as the simulation. Calculating the count separately could cause an off-by-one difference at the boundary.

This should keep the existing time-grid behavior unchanged while making sure setup stays bounded.

wtfPrethiv:
I’ve pushed a fix that stops time-grid generation once it exceeds the precompute cap, so we no longer materialise an oversized grid before falling back to the interpolating provider. I also added focused test coverage for the capped path.

sylvesterkaczmarek (commented):
There is still a correctness issue when callers use fixed_time_step() or jitted_fixed_time_step() with a dt smaller than the configured fixed_dt. _fixed_step() applies min(configured_dt, max_dt), so the simulation can request geometry at a time that is not present in the precomputed grid. With TORAX errors disabled by default, PrecomputedGeometryProvider silently returns the nearest precomputed geometry instead of the correctly interpolated geometry. Could precomputation be disabled for this path, or could off-grid requests fall back to the original interpolating provider?

wtfPrethiv:
Thanks for catching this ! Off-grid requests now fall back to the original interpolating provider, while exact grid matches still use the cache. I added regression tests for shortened steps, repeated calls, and runtime overrides, including JIT coverage

sylvesterkaczmarek (approved):
Rechecked current ddedac82. The off-grid correctness issue I raised is resolved: PrecomputedGeometryProvider now uses cached geometry only for an exact time-grid hit and falls back to the original interpolating provider otherwise. The new regressions cover eager/JIT/vmap evaluation, nearby non-grid times, shortened fixed/jitted steps, repeated calls, and runtime step overrides. This preserves interpolation semantics while retaining the cache on the intended fixed grid. No remaining blocker from my review.

PR #2449, opened 2026-09-10
Add Pylint workflow for Python code analysis

nexoracontrol-ops:
content/actions/how-tos/manage-workflow-runs/approve-runs-from-forks.md

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,23 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
This workflow cannot successfully lint the current TORAX tree as written. The project declares `requires-python = ">=3.12"`, but the matrix only runs 3.8–3.10, and the job installs only `pylint`, not TORAX or its dependencies. That means the proposed check targets unsupported interpreters and will also emit import errors for normal imports such as JAX/Flax rather than useful project diagnostics. Could the matrix use the supported Python versions and install the project/dev extra before invoking pylint (for example, `pip install -e '.[dev]'`)?

